PR #77, opened 2017-11-08
Query parameter/vbg

hitomin21:
Hej, vill få in queryParameter till pre-release.


jacobwod:
Closing after communicating with Hitomi on this.

PR #119, opened 2018-04-07
fixed query variable "z" where parsing 0 caused the value from the co…

hitomin21:
fixed query variable "z" where parsing 0 caused the value from the config to be used due to parseInt(0) is equal to false.Also fixed merge conflict to not open popup in normal snabbsok!


hitomin21:
Fixed a bug in query search IE, created a new pull request for both

PR #131, opened 2018-06-11
Added support to set font family used for print

jacobwod:
Font familj used by mapservice/ when creating PDFS/TIFFS is set by the parameter 'exportFontName' in Web.config. If the parameter is not sat or non-existing, Verdana is used as default font.


jacobwod:
Tested internally, works like a charm. Merging.

PR #242, opened 2019-06-26
Now we're loading Roboto from CDN

jacobwod:
Somehow we missed to add Roboto (can be either from CDN or local, by npm install), even
though we're using it (as it's the default font for Material Design). This way
(<link> in index.html) is the recommended one, according to docs: https://material-ui.com/components/typography/#roboto-font-cdn .


jacobwod:
Absolutely. We should do a lot of things in *admin*.

PR #243, opened 2019-06-27
Important fixes to MapSwitcher:

jacobwod:
- If 'themeMapHeaderCaption' is sat in LayerSwitcher's config, use it as label (i.e. make it possible to customize what the button says)
- If 'instruction' is filled in, use it and display a Tooltip

These settings:

image

Render like this:
image

**BTW: We all should start using <Tooltip> component a lot more in our plugins.**

Signed-off-by: Jacob Wodzyński <jacob.wod@gmail.com>

### DIFF
--- a/new-client/src/components/MapSwitcher.js
+++ b/new-client/src/components/MapSwitcher.js
@@ -1,5 +1,6 @@
 import React from "react";
 import Button from "@material-ui/core/Button";
+import Tooltip from "@material-ui/core/Tooltip";
 import Menu from "@material-ui/core/Menu";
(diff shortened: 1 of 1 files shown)


jacobwod:
This MapSwitcher dropdown is part of the AD integration. I guess it requires activation in *admin* to show up in GUI. Try adding this to your `new-admin/config.json`:

```mapsettings.authentication_active: true```

Though it's strange that you see some of the settings (such as the checkbox to enable MapSwitcher) but not all. Anyway, ensure you have enabled the switch as I mention above and report back if it helped.

PR #249, opened 2019-07-30
Feature/spatial search

sbk-linus:
Added features:

- Draw area to search within
- Select object to search within
- Highlight multiple search results in the map

Upcoming features:

- Filter searchable layers


erikedmark:
All good

PR #266, opened 2019-08-21
Update tabs style

sbk-linus:


new-client/src/components/Window.js
```diff
@@ -96,7 +96,7 @@ const styles = theme => {
     },
     content: {
       flex: "1",
-      overflowY: "auto",
+      overflowY: "scroll",
```
jacobwod: Why should we always want to display vertical scroll bar, even when it's not needed? `auto` should be a better choice here.

PR #302, opened 2019-10-23
Remove all packages from backend. Use "Restore NuGet Packages" in Vis…

maan002:
…ual Studio to get them back.


maan002:
Hej Jacob,

Kan du merga denna. Endast borttag av filer i packages.

PR #310, opened 2019-10-25
fix/drawerVisible-added-to-mapservice

### LINKED ISSUE #309: Let admin select if Drawer should be visible/locked when user first loads the app
Currently Hajk always starts with Drawer hidden. A simple change in AppModel would make it possible to control that setting from admin:
```javascript
  drawerVisible: props.config.mapConfig.map.drawerVisibleAtStart || false,
  drawerPermanent: props.config.mapConfig.map.drawerPermanentAtStart || false
```

Changes to backend and admin will be needed too, naturally.

FredrikEkstrom79:
Added support to set drawerVisible and drawerPermanent status from admin

Fixes #309


jacobwod:
Backend part of #309. Closes #309.

PR #314, opened 2019-10-28
Add options to set window width, length and position in admin

### LINKED ISSUE #295: Add option to set panel width
Add an option in _admin_ to set default width on every panel.

victoraxbom:
This pull request add the possibility to set width, length and position (left or right) to all plugin windows.

Closes #295


new-client/src/plugins/BaseWindowPlugin.js
```diff
@@ -39,8 +39,8 @@ class BaseWindowPlugin extends React.PureComponent {
     this.title = props.options.title || props.custom.title;
     this.description = props.options.description || props.custom.description;
 
-    this.width = props.custom.width || 400;
-    this.height = props.custom.height || "auto";
+    this.width = props.options.width || 400;
+    this.height = props.options.height || "auto";
```
jacobwod: This needs to be modified so that we a) use values from options if specified, b) or fall back to custom (custom per plugin, that is), or finally c) hard-coded values.

Issue #323, opened 2019-11-10
Snabbval (preset) is missing in Hajk3 - new-client


perosoderstrom:
**Describe the bug**
You can configure preset in new-admin but the plugin does not show up in Hajk3 new-client. I guess the plugin is not yet migrated from Hajk2 to Hajk3. There is no folder "preset" in new-client/src/plugins.

**Expected behavior**
When configured and activated in Admin, the plugin Snabbval should show up in client.

jacobwod:
You're right @perosoderstrom, but I'll close this issue since that functionality is already described in #152, and then further expanded upon in #159. Please see if #159 it fulfills your requirements, feel free to comment if something's missing.

jacobwod:
When I think about it tough: we could in the meantime add some info about the plugin not being implemented in `admin` (or hide the configuration). I understand if users get confused the way it works now.

perosoderstrom:
I have created a branch plugin-preset with code for the plugin. Right now it is implemented as an ordinary plugin, whith possible placement in toolbar or left or right (referring to the discussion in issue #159 ).

jacobwod:
This is closed as of #336. The current implementation does however display a Window. Another option would be to do something similar to MapSwitcher, see [this comment](https://github.com/hajkmap/Hajk/issues/159#issuecomment-549362117). It's a rather small fix to convert this `BaseWindowPlugin` to a free floating button below the zoom buttons, but the **question that remains is whether we want it to be free floating or a window based plugin?**

Issue #159, opened 2018-10-12
Plugin: Quick link


jacobwod:
Quick link to given coordinate and zoom level, with a nice label.

Could be placed next to search, in App bar, and used to quickly zoom to a population center in the municipality.

jacobwod:
Already done?

jacobwod:
Similar functionality exists in Informative plugin.

jacobwod:
Is this needed, in a similar way of how it works in Hajk1? Discuss please.

FredrikEkstrom79:
It's a fast way to jump to a location if you don't know where exactly it is.

jacobwod:
Sure, and it's also pretty simple to implement. A list of places (text labels), with each label corresponding to predefined x, y, and z value. 

**The question that remains is rather of visual implementation: how do we make it fit in nicely into current design?** One solution could be similar to how MapSwitcher works: a button on the map that reveals a drop down on click:

![image](https://user-images.githubusercontent.com/110222/68125554-25c9ff00-ff12-11e9-8f8e-7b4e3e3303e1.png)

Another way is using the Speed Dial component, but that will be **limited to 3-6 items**, which might be too little: https://material-ui.com/components/speed-dial/#persistent-action-tooltips.

PR #336, opened 2019-11-22
Plugin preset 

perosoderstrom:
Initial version of Preset plugin referring to issue #323 (and #159).


perosoderstrom:
Note: I missed to undo my change in appConfig.json before committing.

PR #355, opened 2019-12-03
Vt/feature/vtg 741 model communication

sweco-semtto:


new-client/src/plugins/vtsearch/SearchModel.js
```diff
@@ -0,0 +1,99 @@
...
+    this.localObserver.publish("journey-result-begin");
+
+    // Fix parentheses and so on, so that the WKT are geoserver valid.
+    filterOnWkt = this.fixWktForGeoServer(filterOnWkt);
+
+    // Build up the url.
+    let url =
+      "http://sestoas256:8080/geoserver/kartsidan2/ows?service=WFS&version=1.0.0&request=GetFeature&typeName=kartsidan2:journeys&outputFormat=application/json";
+    const viewParams = `&viewparams=filterOnFromDate:${filterOnFromDate};filterOnToDate:${filterOnToDate};filterOnWkt:${filterOnWkt}`;
+    url = url + viewParams;
```
jacobwod: Any changes that you'll lift the `url` param out instead of the hard-coded value? I understand it's early in the process, but we've unfortunately seen stuff like this being forgotten later on (as it "just works" on the developer's machine), and not being fixed upon release.

sweco-semtto: I agree on your comment, I have seen this myself over the years as well. 
In this particular case the reason for this bad code is that a colleague will need this values soon. The risk that this will be forgotten is equals to zero because the next step is to add more functions and fix the urls to not be hard-coded.

jacobwod: OK, let's merge then.

Issue #368, opened 2019-12-10
Add a Rotate Control to Map that rotates the View back to default (North is up)


FatemeHosseini:
**Om man jobbar i mobil, ipad eller touch-skärm, kan rotera i kartan**
Det behövs en plugin (som en knapp) som roterar tillbaka så att norr är upp i kartan. 
OnClick -> norr-upp

![image](https://user-images.githubusercontent.com/36195301/70522667-dc3c8780-1b41-11ea-9c3f-13698123b6bb.png)

**En förbättring är om  ikonen kan rotera när man roterar i kartan så att man ser vilket håll är norr**
![image](https://user-images.githubusercontent.com/36195301/70523120-a21fb580-1b42-11ea-91de-e6fd9c381f2c.png)

jacobwod:
Whoever wants to implement this: try to follow the same approach as in `src/controls/Zoom.js`. There is a built-in OL Contorl, `ol/control/Rotate` that solves this. But simply activating it in `AppModel.js` would render a Component slightly different from our existing buttons (which are built with Material UI). So please see how Zoom is done and try to follow the same approach for Rotate.

jacobwod:
I'm done with this one. Not sure about the icon however. Input on icon choice for this (as well as other Controls and Plugins) is welcome. **Make sure to pick something from https://material-ui.com/components/material-icons/.**

PR #379, opened 2019-12-19
Let the rotation icon always point to north

victoraxbom:
The addition of a rotation button was really good. This adds a small fix that makes the icon always point to the north. Part of #368

### DIFF
--- a/new-client/src/controls/Rotate.js
+++ b/new-client/src/controls/Rotate.js
@@ -57,7 +57,7 @@ const RotateControl = React.memo(props => {
             className={classes.button}
             onClick={rotateNorth}
           >
-            <NavigationIcon />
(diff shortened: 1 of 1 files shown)


jacobwod:
Also, I'm open for changing the icon to something else. I just couldn't find any icon with "N" in it, so the current one was the closest to requested, even if it lacks the direction. The easiest way is to use MUI Icons, https://material-ui.com/components/material-icons/, but we can also **create custom icons from SVGs** as described here: https://material-ui.com/components/icons/#svgicon.

victoraxbom:
No, there doesn't seem to be any perfect icon in the iconpack. Google Maps uses the compass icon, but then it's colored red/white. Doesn't really work with the black and white version as you can't see which side is north. A custom svg icon might be the best way to go then

PR #392, opened 2020-01-22
Utskrift fix/vbg

hitomin21:


madaco-vbg:
We are going to get rid of the formatting changes before merging.

Issue #357, opened 2019-12-03
Clean up settings in admin


jacobwod:
I've recently done some clean ups in `admin`, mainly in `search.jsx` and `infoclick.jsx`. There are still lot of unused options there that only confuse the admin user. We could also remake the structure, add some separators, like those I've added to Infoclick:
![bild](https://user-images.githubusercontent.com/110222/70057359-4a7ec880-15dd-11ea-8e2a-bb24451cde17.png)

Feel free to comment below on which parts of admin you aim to work.

jacobwod:
Info: the separation line is made with CSS and available across `admin`. Use it by simply: 
```html
<div className="separator">Some text here</div>
```

jacobwod:
Is there anything obvious left to do in this issue? Or have we done everything that could be done without major rewrite of how admin and backend work?

jacobwod:
Well done @FredrikEkstrom79, I feel we can close this issue.

PR #397, opened 2020-01-30
Changed inputbox to dropdown with preset option

FredrikEkstrom79:
Change the inputbox for target to a dropdown with preset valid options for all plugins with that input i admin (15 plugins with dummy)
image


jacobwod:
Part of #357

PR #428, opened 2020-02-13
Added options to edit resolutions and origin

FredrikEkstrom79:
Found settings in map-file that where not present in admin, so i added them to mapoptions.
Now you can change resolutions and origin from admin.

image

part of #357


new-admin/src/views/mapoptions.jsx
```diff
@@ -38,6 +38,7 @@ class MapOptions extends Component {
 
   componentDidMount() {
     this.props.model.on("change:mapConfig", e => {
+      console.log(this);
```
jacobwod: Remove logging please.

Issue #425, opened 2020-02-12
Call out: new default WMS service needed


jacobwod:
It seems like the service we've been using has outdated SSL certs, which leads blocked requests:
![image](https://user-images.githubusercontent.com/110222/74320145-68f10500-4d80-11ea-9a75-135e8beaa4d6.png)

I noticed that while preparing the 3.2 release. We should replace all occurrences of `giscloud.se` in our default `layers.json`, before we can do the release. **Anyone with a usable, public WMS service out there? If so, please attach valid `map_1.json` and `layers.json` as a response to this issue and I'll take care of that.

jacobwod:
Regarding this, I'm seriously considering adding OpenStreetMap to default LayerSwitcher (pretty much like white and black backgrounds are available currently). This would solve at least the background layer issue.

jacobwod:
Added OpenStreetMap in 7552cdf719b9aeded1f89f9cd9c9ab7c6a6ab9ab, closing this.

PR #436, opened 2020-02-20
Update configuration files in backend

victoraxbom:
New layer and map files with a basic configuration for the first startup.
You though need to set up the proxy to be able to search

Fixes issue #425


jacobwod:
We'll look into this today. 👍

jacobwod:
It looks fine here, good job @victoraxbom. I will merge it, but I don't know how we should solve the WFS problem. It'd be so nice to have a working search box, but that would require an open, meaningful WFS service in the same extent as our default map is.

victoraxbom:
@jacobwod Yes, it would be perfect if the search would work right of the box. There is an open WFS included in the config files which works fine, the problem is CORS, which the proxy solves by adding the cors headers to the response from the wfs server. Don't know if anyone have another solution? 
Maybe the proxy could be pre-configured in the release

jacobwod:
Ah I see. I don't understand why they provide an "open" WFS service but don't send the `Access-Control-Allow-Origin: *` header.

jacobwod:
I would be cautions with adding the proxy component into the default configuration however (even if it'd solve this current problem). The reason is that adding proxy adds complexity to the system and introduces another point of possible failure. We've previously had problems with the included .NET proxy and have disabled it since, in favour of NGINX's proxy module which is more robust. 

**To sum up: it should be up to the admin to decide whether to use proxy or not, but by default we should keep it simple.**

PR #459, opened 2020-03-10
Utskrift fix/vbg

hitomin21:
configuration "exportScaleberSetting" has added in web.config to adjust scalebar length for each size/orientation/scale.


jacobwod:
@hitomin21 Feel free to review it internally with your colleagues. I'm not going to look at this as it's aimed for the 2.x branch.

maan002:
OK

Issue #435, opened 2020-02-20
Error on building mapservice "Microsoft.net.compiler not supported"


hitomin21:
We have problems building mapservice for HAJK3 on branch "master" and commit "723d5f09b3cabbc4b915418d8b732900bf2674f7 [723d5f0]". 
I, Mattias and Emma get the following error when attempting to build.

![bild](https://user-images.githubusercontent.com/22234673/74946326-63876080-53f9-11ea-98a9-071ee43e18b4.png)

We've tried with rebuild and re-downloading the repo without any luck. Does anyone have a solution? We're using the latest version of Visual studio (2019).
Jacob, I assigned this to you since you pushed the most recent commit regarding NuGet.

jacobwod:
@FredrikEkstrom79 Do you have any problems building the .NET application? Can you provide details upon which version of VS etc you're using?

FredrikEkstrom79:
It's a bug in VS 2019, even if compiler version is above 16.3 it will fail the version check.
https://developercommunity.visualstudio.com/content/problem/842373/microsoftnetcompilers-is-only-supported-on-msbuild.html

As the post in the middle suggest, try comment out the "Error" part in this file
**Hajk\backend\packages\Microsoft.Net.Compilers.3.4.0\build\Microsoft.Net.Compilers.props**
![image](https://user-images.githubusercontent.com/50202826/75329702-61e9ec80-5880-11ea-89a0-ec8363286409.png)

Now you should be able to build the project

jacobwod:
Good info @FredrikEkstrom79 . Closing this issue as it's a bug outside of our control and a workaround is now provided.

FredrikEkstrom79:
I better solution is to rename MSBuildToolVersion to MSBuildVersion.
This will be fixed in the 3.5.0 release of this NuGet package, i have tested 3.5.0.beta3-final and i works fine

https://github.com/dotnet/roslyn/pull/40349/commits/1b30a771a9428f49049f3a2f8b04ed7c724c04a2

![image](https://user-images.githubusercontent.com/50202826/76391020-d0de3f80-636e-11ea-820d-9d30de5b5bf5.png)

PR #470, opened 2020-03-20
Updated to final 3.5.0

FredrikEkstrom79:
Updated Microsot.Net.Compiles to final stable version 3.5.0.


jacobwod:
Related to #435 and #460. This should finally close this for good.

PR #499, opened 2020-04-30
Add CQL filters for WMS layers

jacobwod:


jacobwod:
Pretty much done. Here's how it work.

### No CQL filter
Skärmavbild 2020-04-30 kl  12 37 13
All features are shown.

### CQL filter active
Skärmavbild 2020-04-30 kl  12 36 57
With the CQL expression `postort='Harplinge'` we've narrowed down the results.

### Integration with Anchor plugin
Skärmavbild 2020-04-30 kl  12 40 17
Each layer can have it's own CQL filter active. The CQL filters are sharable via the Anchor plugin where they are exposed as the `f`-parameter. 

### Works with `clean` mode
Because the `f`-parameter is exposed, it is also available on load. When `AppModel` parses the query string and finds CQL filters, they are applied for the specified layers. 

**This makes it possible to create an embeddable map, zoomed in on a certain feature and only showing that one - and this is the sole motivation behind this feature.** Now, we can use the same layer of for example schools, but hide all schools except one, which we zoom into and show its location on a clean, embeddable map.

PR #512, opened 2020-05-12
Kir fix/vbg

hitomin21:


madaco-vbg:
attempting to merge to wrong branch

PR #536, opened 2020-06-09
added "stäng" and "Tyck till igen" buttons. "stäng" close the tool-wi…

hitomin21:


jacobwod (changes requested):
Please use English in the code (whenever it's naming objects, functions, comments etc).

new-client/src/plugins/collector/collector.js
```diff
@@ -28,6 +28,8 @@ class Collector extends React.Component {
         serviceConfig: this.serviceConfig
       }
     });
+
+    this.collectorModel.app = props.app;
```
jacobwod, line 32: What's the point of this? You already send the app as prop when model is instantiated (line 23). 
```suggestion
// Remove line
```

new-client/src/plugins/collector/components/Page.js
```diff
@@ -573,8 +573,47 @@ class Page extends Component {
...
+        variant="outlined"
+        color="primary"
+        className={classes.buttonRight}
+        onClick={() => {
+          this.props.model.observer.publish("abort");
+        }}
+      >
+        Tyck till igen!
+      </Button>
+    );
```
jacobwod, line 587: `collectAgainButton` and `okButton` are virtually the same, except for the label.

Just use one of them, read the value of `this.props.options.collectAgain` inside it and set label accordingly.

jacobwod (changes requested):
See comments in files.

new-client/src/plugins/collector/collector.js
```diff
@@ -28,8 +28,7 @@ class Collector extends React.Component {
         serviceConfig: this.serviceConfig
       }
     });
-
-    this.collectorModel.app = props.app;
+// Remove line
```
jacobwod, line 31:
```suggestion
```

jacobwod (changes requested):
The idea was that you actually remove the line, not add my comment… 😄

new-client/src/plugins/collector/collector.js
hitomin21: I noticed and fixed it in another commit :)

new-client/src/plugins/collector/components/Page.js
```diff
@@ -568,13 +573,39 @@ class Page extends Component {
         onClick={() => {
           this.props.model.observer.publish("abort");
         }}
+      >
+        {okButtonText}
```
jacobwod, line 577: Drop the `okButtonText` variable and just do a conditional render like this:
`this.props.options.collectAgain ? "Tyck till igen : "Stäng"`

jacobwod (changes requested):

jacobwod (approved):

PR #626, opened 2020-11-26
Fix/drawer content configuration at start

sweco-secoma:
1. Fix problem that the drawer can be open with empty content if the drawer is configured to be open on start but there is no active content in local storage. There is an option in config to set the active drawer at the beginning. The drawer will default to plugins if this is not used or empty. The correct drawer button will also be activated. 

2. Change so that the draw is not set to permanent when opening the map on mobile.


jacobwod (approved):
Looks good!

PR #629, opened 2020-12-08
Mobile start configuration changes

sweco-secoma:
Added a separate config property to set whether the drawer should be open on start on mobile. This is to give the option of choosing to have the drawer be open at start if opening on a desktop, but not have the drawer open if opening on a mobile. 

Also simplified the drawer logic relating to mobile settings. It first checks if on a mobile, and if so, deals with that - otherwise it runs the original drawer logic as normal. 

**Question:**
For this change to be effective - there probably should also be a mobile specific configuration for whether the layerswitcher itself is visible on start. The simplest way is to do it in exactly the same way as have done now, to add visibleAtStartMobile in the layerswitcher tool configuration. 

I'm conscious of filling up the configuration with lots of extra mobile properties - would something like the following be better? 

`visibleAtStart: {
  "desktop": true,
  "mobile": false 
}`

It keeps it together a bit more, but comes with a few downsides - the logic changes needed wouldn't work with any previous config for example. 

Any thoughts?

If no objections, will implement the normal way (with a visibleAtStartMobile for layerswitcher) for the moment.

### DIFF
--- a/backend/mapservice/App_Data/map_1.json
+++ b/backend/mapservice/App_Data/map_1.json
@@ -156,7 +156,7 @@
           }
         ],
         "active": true,
-        "visibleAtStart": false,
+        "visibleAtStart": true,
         "backgroundSwitcherBlack": false,
         "backgroundSwitcherWhite": true,
         "enableOSM": true,
@@ -167,7 +167,6 @@
         "height": 0,
         "title": "",
         "description": "",
-        "enableTransparencySlider": true,
         "dropdownThemeMaps": true,
         "themeMapHeaderCaption": "Standardkartan",
         "instruction": "",
--- a/backend/mapservice/App_Data/map_1.json
+++ b/backend/mapservice/App_Data/map_1.json
@@ -437,6 +437,7 @@
     "mapcleaner": true,
     "drawerVisible": true,
     "drawerPermanent": true,
+    "drawerVisibleMobile": true,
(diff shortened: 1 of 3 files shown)


jacobwod (approved):

jacobwod:
Good idea but it requires changes in admin that might be difficult in the current solution. Go on with `visibleAtStartMobile`.

sweco-secoma:
@jacobwod 

Have made some updates since you approved if you want to check further. 

Added visibleAtStartMobile for layerswitcher - changed setting of visible state in baseWindowPlugin for this. 
changed to use getMobile instead of isMobile

jacobwod (commented):
Are you aware of the change (`isMobile` -> `getIsMobile`)? The last one only checks for `window.innerWidth` so we can't be sure that this will cache all handheld devices (a lot of high-res displays in horizontal layout will have a width > 600.

jacobwod:
I know that `getIsMobile` is used in `Window` currently, but I'm not a fan. Perhaps we should re-write that part too, utilize [MUI's breakpoints](https://material-ui.com/customization/breakpoints/) or the [Hidden component](https://material-ui.com/components/hidden/) where applicable?

sweco-secoma:
I was aware of it but had another look after reading your comment and saw that the isMobile check is anyway doing the width check that getIsMobile does, so i've changed it back.

Issue #628, opened 2020-12-07
Error loading WFS Vector layer


persoderstrom:
**Error message when loading WFS Vector layer**

_Cannot read property 'getAxisOrientation' of null_

Vector/WFS has worked before. 

**To Reproduce**
1. Add a Vector layer (Point) with WFS format in admin
2. Add the layer to a map
3. Open the map and select the layer
4. No features displays. See error in Console

**Screenshot**
Error in Console

![Skärmbild 2020-12-07 165521](https://user-images.githubusercontent.com/1893510/101373497-4bf86600-38ad-11eb-9190-ea6b4e016ee8.png)

persoderstrom:
This might be caused by changes in "Feature/592 snap interaction"

jacobwod:
I haven't tried this with a pure Points layer (my tables hold various types of geometries), but I can understand that it would be impossible to get an orientation out of a point. 

However, the error indicates that you don't even have a geometry, but rather a null object, which naturally lacks any properties, including getAxisOrientation. 

A quick search tells me that you might be missing the "duplicate" projections definitions in your map config. It seems that the GML parser reads "the other" format, [see this thread](https://stackoverflow.com/questions/36087622/ol3-polygon-wfs-feature-search-axis-orientation-undefined). E.g. for 3006 you should have these two:
```
    {
      "code": "EPSG:3006",
      "definition": "+proj=utm +zone=33 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs",
      "extent": [
        181896.33,
        6101648.07,
        864416,
        7689478.3
      ],
      "units": null
    },
    {
      "code": "http://www.opengis.net/gml/srs/epsg.xml#3006",
      "definition": "+proj=utm +zone=33 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs",
      "extent": [
        181896.33,
        6101648.07,
        864416,
        7689478.3
      ],
      "units": null
    },
```

perosoderstrom:
Problem solved. Duplicate projections was missing. 
Thanks, @jacobwod !

persoderstrom:
In default map_1 the only projection that is duplicated is EPSG:3006.

jacobwod:
Feel free to fix and send PR.

PR #631, opened 2020-12-08
Removed all projections (except 3006) from default map config. The rest can be found in templates/projections.template.

persoderstrom:
Solves problem with Vector/WFS, issue #628


jacobwod (approved):

jacobwod (changes requested):
Let's await the changes which we spoke about on today's meeting, I'll merge it then. 👍

jacobwod (approved):

Issue #482, opened 2020-03-27
Improve Coordinates plugin - allow users to enter coordinate


jacobwod:
Currently the Coordinate plugin allows user to click anywhere on the map and displays coordinates for the click point in different (admin-defined) projections:
<img width="408" alt="Skärmavbild 2020-03-27 kl  08 41 44" src="https://user-images.githubusercontent.com/110222/77733432-2a00d100-7007-11ea-803f-2dff5b6c1b5e.png">

There are cases where user needs to enter coordinates by hand into the input boxes and transform between different projections. **Therefore it would be a great enhancement if the input boxes, that currently are read-only, allowed (well…) input too.**

Part 2: We've also had requests about showing the coordinate `onMouseMove` and not just `onMouseClick`.

jacobwod:
@sbk-linus Can you estimate how much work would be needed for this?

hitomin21:
I will update the coordinates tool to have the same functionality as in Hajk2 so I will add so the user can input values.

I have added so it is possible to modify the input fields and marker as well as other fields are updated accordingly.

I would also like to add buttons at the bottom:
Zooma -> Centers on written coordinate and zooms in
Panorera -> Centers on written coordinate
Clear -> Removes the marker

Please let me know if anyone has feedback regarding this.

jacobwod:
@hitomin21 Sounds good, please remember to develop this feature in a separate branch, and send a PR when you're done about merging it into `master`.

hitomin21:
I am working with this in the branch https://github.com/hajkmap/Hajk/tree/updatecoord/vbg

hitomin21:
I will add some more improvements for this tool and corresponding configurations.

Users have reported that sometimes several digits are being deleted although only one should have been deleted.

Add a my location button to allow users to see the coordinates of their current position

Buggfix on number of decimals

Number formatting 1000 -> 1 000

The default view shows all the input fields directly so the user does not have to click somewhere in order to specify a coordinate

A button for reseting the fields

jacobwod:
Go on @hitomin21. Make sure to keep your feature branch up-to-date with latest development in `develop`. When you're done, make a PR to `develop`.

hitomin21:
Working on branch 
https://github.com/hajkmap/Hajk/compare/feature/482-improved-coordinates

hitomin21:
Merged into develop

jacobwod:
[...]

PR #639, opened 2021-01-04
Updatecoord/vbg

hitomin21:
Coordinates tool now allows users to modify the coordinates in the textboxes and the feature on the map will be moved accordingly. Two buttons have been added.

One for zooming in, and panning, on the selected location and one for only panning to the selected location.

@persoderstrom  @perosoderstrom Can you take a look?

The version in the pull request looks like this:
bild

closes https://github.com/hajkmap/Hajk/issues/482

Note that I did not remove addInteraction and removeInteraction (it looks like this in the diff) but code was added in the middle.


persoderstrom (commented):
@hitomin21 It's working good!
Maybe you should add a friendly message if a user types in NaN by mistake (or deliberately as I did). If you do and try to Zoom or Pan you get an error page. 
Alternatively disable Zoom/Pan if the input filelds are not numbers.

hitomin21:
I have fixed so there is no error page when non-finite numbers are inserted so users will not be presented with the error page that you described.

There is one remaining problem, if a user writes a very large number and clicks zoom or pan, the openlayers function that would zoom or pan hangs. No error page is shown and the page freezes until the user refreshes. Perhaps this is ok? I think adding extent checking would require changes to the tool configuration and therefore an admin update. I am not sure it is worth it in this case.

Per: What do you think?

persoderstrom:
@hitomin21 Good!
One thing that might happen is that a user edit a coordinate and by mistake delete the decimal point, which increase the number by a factor of 1000. I tested and that does not hang the page. You have to write numers greater than 10e20 to hang the map. So, if there is not a simple way to get the map extent, leave it as it is.

hitomin21:
I pushed the error handling code now, please take a look again.

persoderstrom:
The error handling works ok. 
But when you have edited a coordinate and panned and then close the tool you get an error page.

jacobwod (changes requested):
Two things:
- You seem to use old Prettier as you're commits remove all trailing commas. Please ensure you're up-to-date (`npm up` should be enough but you can also do `rm node_modules && git pull && npm i`).
- The error handling you added doesn't seem to do any error recovery whatsoever – you just return from the `catch`.

hitomin21:
I have fixed the error that you mentioned Per and ran Prettifier on the code.

Before I push it, @jacobwod please see below.

Regarding the error recovery, when invalid coordinates are entered into the box I thought that the user might be in the middle of modifying them and it is best not to show an alert popup with a message. Do you think that would be preferred? The function updateTransforms calculates the new transformed coordinates but since the inputted data is invalid it is best just to quit the function so no new faulty transferred data is written to the gui.

jacobwod:
@hitomin21, I can't see your code until you've pushed it to this branch (and there's no harm doing it as the PR isn't merged yet). 

Also, I don't see why you use `try/catch` the way you do: that should be used for unexpected errors and recovery from them. In this case you expect that almost every keypress will result in an incomplete value that can't be parsed, and this throws an error, of course. But it shouldn't go that far: you should reject early in the `handleInput` method, and continue parsing only when some specified conditions of the input value are fulfilled (for example the input has a specific length, contains `.`, etc -  use a [`/someRegex/.test()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test#examples) to check if the condition is fulfilled).

hitomin21:
I've pushed now.

There is verification in handleInput, the input will be parsed as floats and are verified that they are not NaN and that they are finite.

The remaining problem is mainly if the inputted numbers are in range for the transform, which would cause openlayers to throw an exception. This is what I mentioned above regarding that we would likely have to add an extent for each configured transform for the tool.

jacobwod (changes requested):
@hitomin21, I still feel that the `handleInput`, which will run a lot when user is typing, is a bit too heavy, and therefore resource-intensive. 

The way you grab values (by traversing the DOM and looking for parent's parent) is also error-prone, perhaps there is some other way you could extract those values?

jacobwod:
Perhaps you could re-think the way that validation is done too? There are great helpers in MUI, please take time to ensure you're using them, e.g. in the TextField, to handle validation of user input. Everything is well-documented [right here](https://material-ui.com/components/text-fields/#validation).

hitomin21:
I added so we use the error attributes of MUI to show the error messages to the user. The field becomes red and there is message to the user that it should input a valid number now.

I profiled the JS with dev tools and it does not use so much performance when I keep changing the numbers a lot so I think we should be ok performance-wise. I put the input validation at the top of the code so errors will be detected directly instead of after the DOM was parsed although that would only help when users write incorrect numbers of course.

Regarding using the DOM, can we accept that if we modify the structure of the tool that we also has to updated how the code finds the sibling to the field that was modified?

jacobwod:
Please push the code to the branch so that I can see the changes you're talking about. Thank you.

hitomin21:
I pushed the code now, sorry I forgot to click it yesterday. I merged master into the branch but I want to test it some more, I will write here after I have tested it.

hitomin21:
I have tested it and it works with the current master so please review it again @jacobwod

jacobwod (changes requested):
Hi @hitomin21! A quick look at the `handleInput` tells me that it should be better now - you seem to do an early return when invalid values occur. I haven't tested it yet however.

Still some minor improvements that can be made: please replace `var` with either `const` or `let` depending on situation. (I usually go with `const` unless I see that I need to re-assign a variable - in that case I just change it to `let`.)

hitomin21:
I modified all the var to let and const now, thanks for that.

sweco-sesqqq (approved):
I just ran the code locally and everything seems to be working.

There seems to be some stability issues connected to handle input still, look in to it if you want to! :)

PR #681, opened 2021-02-23
Vt/core changes

sweco-semtto:
Västtrafik's own plugin.


sweco-sesqqq:
A quick overview of this pr shows several problems that must be handled:

- Wrong encoding in README
- Several changes in new-_backend_, i would guess maybey you ran prettier on new-backend? This must be reverted.
- It looks like DocumentSearchModel.js and BackgroundSwitcher.js has been renamed. It seems like the filename is unchanged, but case has changed.

PR #693, opened 2021-03-11
Feature/657 fetch options

jesade-vbg:

PR #694, opened 2021-03-11
Feature/657 fetch options

jacobwod:
Let's review it one last time!


new-admin/jsconfig.json
```diff
@@ -0,0 +1,5 @@
+{
+  "compilerOptions": {
+    "baseUrl": "src"
+  }
```
jacobwod, line 4: Just curious: what does this do and why was it needed?

new-admin/src/index.js
```diff
@@ -18,11 +18,19 @@ import Backbone from "backbone";
 import { Router } from "backbone";
 import ApplicationView from "./views/application.jsx";
 import ApplicationModel from "./models/application.js";
-
+import {
+  initHFetch,
+  hfetch,
+  initFetchWrapper,
+  wrapJqueryAjax,
+} from "./utils/FetchWrapper";
```
jacobwod, line 26: I haven't noticed until now, but it seem like we have `utils/FetchWrapper` in both `new-admin` and `new-client`. Can't they be consolidated, somehow? I now that we haven't shared code between the different applications before, and perhaps we can wait. But it would be nice not to duplicate and put shared components somewhere global. (On second thought: I guess we should create an issue so this is not forgotten, and just go on as-is. Opinions, @sweco-sesqqq @jesade-vbg?)

new-client/prebuild.js
```diff
@@ -0,0 +1,59 @@
+const { execSync } = require("child_process");
+const fs = require("fs");
+
+const envLocalFile = ".env.local";
```
jacobwod, line 4: Can you please add some defaults to `.env`, so admins can figure out the key names without looking inside the code? I guess that we can set our own values in `.env.local`, but defaults would be nice too.

jacobwod (commented):
Great, just a couple of things I'd like to discuss, see line comments!

new-admin/jsconfig.json
jesade-vbg: It solved my problem (something with webpack maybe) that I could not import anything from a subfolder in index.js.
For example when testing: This won't get resolved in index.js  >   import { arraySort } from "./utils/ArraySort";
I tried, but could not get it to work without the compilerOptions. Maybe I'm missing something.

new-admin/src/index.js
jesade-vbg: Yes, this is not optimal. Maybe we could adress this "shared code" issue when creating the new admin? Or separate issue as you wrote.

new-client/prebuild.js
jesade-vbg: Yes, I've now added defaults. Also moved the npm vars to .env as these never change in prebuild.js

jacobwod:
Closing this as it's been accidentally created. See #693.

PR #695, opened 2021-03-11
Added consistent height for parent-children in drawer to fix scroll iOS

sweco-seqtob:
Minor fix
Setting height for all tree of containers in drawer to avoid height set to auto. 
This fixes problem with non-working automatic scrolling in all iOS-browsers.


jacobwod (approved):

sweco-sesqqq (approved):
Looks good!

Issue #615, opened 2020-11-08
Collector: Allow users to insert multiple geometries in the same feedback


hitomin21:
Users are limited to only providing one geometry per feedback. So if an user wants to give the same feedback about n areas, it needs to create n feedbacks.

Currently, the answers are inserted into a geometry which is placed in the database. Instead, we should create WKTs from the geometries that the user placed for the feedback. We can then have multiple questions where the user can specify geometries instead of only one.

Putting several separate feedback features in the database would create an issue if linking these back when displaying  the answers to the administrators. I think the WKT approach where everything is stored in one feature is preferred.

perosoderstrom:
Hi Hitomi!
We use the feedback information in the database in a task management system, so the WKT solution will cause problems for us. 
We also think it is generally best so store information in a database rather than in files. 
A solution might be to store the feedback information in one table, and the geometry/geometries in another table and joining them by common id:s in a view.

hitomin21:
Hi,

I know Mattias contacted you Per but I will write here so everyone can see the answer.

The new functionality has to be enabled in the config. Otherwise the tool will continue to work as it did previously. The idea is that WKT should be used to put the WKT strings into different columns in the same database row. No information is stored in files in the solution I implemented.

persoderstrom:
Hi!
Good @hitomin21! 
Now I know how the purpose of WKT and that it is configurable.

hitomin21:
Feature has been merged into develop

PR #706, opened 2021-03-16
Feature/615 multiple features collector

hitomin21:
Support for adding multiple geometries for an answer in the Collector as well as adding geometry answers to multiple questions.

The new features can be enabled in admin using the WKT option, if this is disabled then the collector will work as before.

Linked issue:
https://github.com/hajkmap/Hajk/issues/615


new-client/src/plugins/Collector/CollectorModel.js
```diff
@@ -26,6 +26,10 @@ class CollectorModel {
     this.filty = false;
     this.saving = false;
     this.geometryName = "geom";
+    this.wkt =
+      typeof settings.options.wkt === "undefined"
+        ? false
+        : settings.options.wkt;
```
sweco-sesqqq, line 32: I would suggest the nullish coalescing operator here instead (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator). (Just a suggestion, everything will work fine as is).

new-client/src/plugins/Collector/components/Toolbar.js
```diff
@@ -37,6 +38,24 @@ class Toolbar extends Component {
...
+      this.props.model.vectorSource.clear();
+
+      try {
+        const format = new WKT();
+        let features = format.readFeaturesFromText(
+          this.props.model.formValues[this.props.field]
+        );
+        features.forEach((feature) => {
+          this.props.model.vectorSource.addFeature(feature);
+        });
```
sweco-sesqqq, line 53: The readFeaturesFromText is a protected method and should not be used i think. See, row 625 in new-client/node_modules/ol/format/WKT.js. Maybey there is a better method to use? If not, i will still accept the change since this is a stand-alone plugin. 

I think we can avoid the try-catch by making sure features is an array of features (or empty) before trying to map over it?

new-client/src/plugins/Collector/CollectorModel.js
hitomin21: I switched to using the nullish operator, thank you!

new-client/src/plugins/Collector/components/Toolbar.js
hitomin21: Now using readFeatures instead and restructured to avoid the try-catch as well.

sweco-sesqqq (approved):
Looks good!

Issue #715, opened 2021-03-24
Can't input text on Mobile


jesade-vbg:
When I try to input text in a textfield the keyboard pops up then it closes directly after, the input loses focus.

The only text-input that I can find that works is the search-input.

Bug found in:
* Print > input title
* Bookmark > input name
* Draw Circle > input radius.
* Color picker > input hex-color or RBGA
* etc etc etc.

Bug can be found in:
https://karta.halmstad.se
and
https://hajk-demo.netlify.app/

Thoughts:
Maybe globally registered document.focus or input.blur somewhere in code where it should be locally handled.
I think this points to the search input as it's the only input I can get working. But I havn't investigated any further.

Browser:
Chrome (Latest) on Android.

jesade-vbg:
The issue seems to be isolated to Android. Tested with Safari and Chrome on iOS with no related issues.

jesade-vbg:
https://user-images.githubusercontent.com/73874338/112308390-5e52b000-8ca2-11eb-8942-8c4d73c9fb29.mov

sweco-sesqqq:
I have to get an android phone... Could you look into this since you've got an android?

jesade-vbg:
I've been investigating the better part of the day.
I think I've found the bad guy, but I'm not sure yet how to solve it and what consequence it has.

When I remove setTimeout and the focus() function (link below) it executes, the problem goes away on my Android phone.

https://github.com/hajkmap/Hajk/blob/dd22c06f0877890b543d1bf4c0b3b385f94d33e9/new-client/src/components/Window.js#L184

sweco-sesqqq:
Nice work! @sweco-seqtob could you explain what would happen if we removed that snippet? Was the focus added so that screen readers would read in correct order?

jacobwod:
Good find @jesade-vbg! 

We still need to find another way of focusing to make @sweco-seqtob happy, though. ;) Again, what was the reason that this focusing was added?

jacobwod:
Well done, it's in `develop` now and ready for tests!

PR #716, opened 2021-03-24
Vt/core changes 4

sweco-semtto:
Added dynamic height for the window class and the vtsearch plugin that renders elsewhere.


new-client/src/components/App.js
```diff
@@ -485,7 +485,9 @@ class App extends React.PureComponent {
 
   renderSearchComponent() {
     // FIXME: We should get config from somewhere else now when Search is part of Core
-    if (this.appModel.plugins.search) {
+    const renderSearchComponent = !this.appModel.plugins.search.options
+      .renderElsewhere;
+    if (renderSearchComponent && this.appModel.plugins.search) {
```
sweco-sesqqq, line 490: Did you test this with a config without a search-component? I think this.appModel.plugins.search.options will be undefined in that case, causing a crash?

new-client/src/components/Window.js
```diff
@@ -175,17 +178,9 @@ class Window extends React.PureComponent {
...
 
-  componentDidUpdate = (prevProps, prevState) => {
-    if (this.props.open) {
-      //This is ugly but there is a timing problem further down somewhere (i suppose?).
-      //componentDidUpdate is run before the render is actually fully completed and the DOM is ready
-      setTimeout(() => {
-        this.windowRef.current.focus();
-      }, 200);
-    }
-  };
```
sweco-sesqqq, line 188: I don't understand why this is removed? I might be missing something? The snipped was added just 3 weeks ago so that hajk could comply with WCAG. Maybey you've solved it some other way?

new-client/src/components/Window.js
```diff
@@ -281,9 +276,11 @@ class Window extends React.PureComponent {
   }
 
   close = (e) => {
-    const { onClose } = this.props;
+    const { onClose, globalObserver, title } = this.props;
     this.latestWidth = this.rnd.getSelfElement().clientWidth;
     if (onClose) onClose();
+
+    globalObserver.publish("window-close", title);
```
sweco-sesqqq, line 283: Could be nice with a better naming here? Could core.closeWindow be a better idea? This would comply with the other event names.

new-client/src/components/Window.js
```diff
@@ -364,10 +370,12 @@ class Window extends React.PureComponent {
     // Run callbacks
     typeof onMaximize === "function" && onMaximize();
     typeof onResize === "function" && onResize();
+
+    globalObserver.publish("window-maximize", title);
```
sweco-sesqqq, line 374: See previous comment

new-client/src/components/Window.js
```diff
@@ -339,13 +336,22 @@ class Window extends React.PureComponent {
...
+      onResize,
+      allowMaximizedWindow,
+      title,
+    } = this.props;
 
     getIsMobile() && this.rnd.updatePosition({ y: 0 });
 
     switch (this.state.mode) {
       case "minimized":
         // Enlarge back to "window" mode
```
sweco-sesqqq, line 351: This comment isn't correct anymore?

new-client/src/components/Window.js
```diff
@@ -384,6 +392,8 @@ class Window extends React.PureComponent {
     // Run callbacks
     typeof onMinimize === "function" && onMinimize();
     typeof onResize === "function" && onResize();
+
+    globalObserver.publish("window-minimize", title);
```
sweco-sesqqq, line 396: See previous

sweco-sesqqq (changes requested):

new-client/src/components/App.js
sweco-semtto: Well spotted. I will update.

new-client/src/components/Window.js
sweco-semtto: My bad. I will update.

new-client/src/components/Window.js
sweco-semtto: Like your suggestion. I will update.

new-client/src/components/Window.js
sweco-semtto: Like your suggestion. I will update.

new-client/src/components/Window.js
sweco-semtto: Like your suggestion. I will update.

sweco-semtto:
Nice comments, I fully agree to all of them. A new updated pull request will appear shortly.

new-client/src/components/Window.js
jesade-vbg: @sweco-sesqqq This is interresting as the settimeout seams to be related to the Android bug #715.

sweco-sesqqq: What a coincidence! (Or maybey you've fixed #715 for Västtrafik Martin?)

I guess the android bug is more substantial than whatever this snippet fixes. Maybey @sweco-seqtob or @jacobwod got some input on this?

sweco-seqtob: This fix is needed to make window focused when opened (implemented due to WCAG requirements). 
It can be removed as long as we can achieve the same functionality in another way?

Problem is that we don't have a callback for when the windows are actually done being painted in the browser (only the postrender method componentDidUpdate). Been trying to use requestAnimationFrame and other functions to make it work.

There shouldn't be any more code running when the setTimeoutCallback is coming back so it would not take more than 200ms and the user wouldn't be able to focus a field in that time, but at the same time it sounds like this is the problem :) (#715)?

Dirty fix  might be to check if it is documenthandler that is opened?

Feel free to refactor if anyone can think of something better :)

new-client/src/components/Window.js
sweco-semtto: Can't find the comment. Wrong lines of code highlighted?

new-client/src/components/Window.js
jacobwod: I propose the following:
1. Martin merges the current `develop` into his branch and makes the PR
2. After the merge, `develop` will still contain Tobias fix for window focus. 
3. Someone, perhaps @jesade-vbg (as you seem to have a test rig ready with an Android phone) makes a local copy of `develop` where he _removes_ the code in `Window.js`, `componentDidUpdate()`, and tests again to see if it changes anything.

Depending on the results of 3. we will either a) remove the fix in `Window.js` (and look for another solution to Tobias' problem!), or b) leave the code as-is (if it doesn't change anything with Jesper's bug) and start looking for a solution to #715 somewhere else.

new-client/src/components/Window.js
sweco-sesqqq: Woops, i think it was "// Enlarge back to "window" mode" i was on about

new-client/src/components/Window.js
jacobwod: OK, I just saw that @jesade-vbg already did the test and it turns out that the related snippet _is_ causing this bug. 

But let's keep the issues apart: @sweco-semtto merges whatever is in `develop` into his branch, at time of merge, while we investigate and find a solution for #715 (and continue the discussion over there). OK?

sweco-semtto: Great!

new-client/src/components/Window.js
sweco-semtto: Unfortunately, I don't quite follow you. What's wrong with the current comment?

sweco-sesqqq: I don't really remember what i meant hehe. Just leave it as is! :)

sweco-semtto: Great! :)

sweco-sesqqq (approved):

Issue #711, opened 2021-03-18
Drawer empty in certain conditions


jacobwod:
I think we have a new issue related to #626. Imaging this:

- We have map_1 where DocumentHandler exists as plugin. User can play around, perhaps lock the Drawer in permanent state and show the DocumentHandler menu. **Those actions will lead to `drawerPermanent:"true"` and `activeDrawerContent: "menu"` being saved to Local Storage.**
- We have another map, `map_2` that _doesn't_ have DocumentHandler. 
- When `map_2` is loaded, Hajk checks if Local Storage has a saved values for `drawerPermanent`. Hey, it turns out it has (rembember, it was saved in `map_1`)! OK, so we will start with Drawer visible. But, what should be shown? Let's check  `activeDrawerContent`! Aha… something called "menu", fair enough.
**- Too bad that we don't have any DocumentHandler in this map, so nothing will take care of filling Drawer with "menu" items.**
- And we end up with this:
![Skärmavbild 2021-03-18 kl  14 50 18](https://user-images.githubusercontent.com/110222/111637997-3541a380-87fa-11eb-8da3-42310ede3356.png)

Can @sweco-secoma or @sweco-seqtob take this to your JIRA?

sweco-seqtob:
Linked a PR to resolve this issue. 
Need some help to verify that it works in all scenarios, because there's a lot of scenarios that can arise from different map setups

sweco-sesqqq:
Just merged it to develop. Testing asap @sweco-seqtob

jacobwod:
It's in develop now, I'll take a closer look into it soon.

jacobwod:
Seems to work fine! 👍

PR #717, opened 2021-03-24
Resolve empty drawers showing under certain conditions

sweco-seqtob:
Resolve empty drawers showing under certain conditions see #711.

- Drawer is now not shown if we don't have either some customDrawerContent or plugins to render in the drawer.
- Fallback is always plugins as long as the mapconfig actually has any tools active

Need help with testing because there are a lot of conditions arising with different mapconfigs


sweco-sesqqq (approved):
I like it 10/10

PR #723, opened 2021-03-26
Feature/669 new featureinfo

### LINKED ISSUE #669: FeatureInfo: Render Markdown into React components, add support for conditional rendering
# Problem
The current implementation has two major flaws:
1. The markdown can not contain any [tags]. This is due to the _shortcode_ feature in FeatureInfo that basically sees everything between [ and ] as a tag. This conflicts with Markdown's Links and Images syntax.
2. The resulting HTML is plain HTML, whereas we'd like to use MUI components on some ocassions. For example, a table should not parse into `<table>` but rather the MUI Table Component, to ensure correct rendering, dark mode support etc.

# Suggestions
1. **I will initially remove the _shortcode_ functionality.** We could look into making something similar, but it must not conflict with Markdown syntax.
2. I'll see if our Markdown parser can be extended to render MUI Components. A quick search gave [this promising result](https://marked.js.org/using_pro). Update: using `react-markdown` makes this possible.
3. Implement some kind of conditional rendering. It could be used e.g not to render empty fields.

# Currently implemented
- Fixed problem 1: Markdown can use all []-tags.
- Fixed problem 2: We return React components
- Added support for suggestion 3, see below.

## Example of conditional rendering
```
<if {foo}>Will show if 'foo' is truthy</if>

<if {foo}=bar>Will show if 'foo' evaluates to 'bar'</if>
```

jacobwod:
Closes #669


new-admin/src/views/tools/infoclick.jsx
```diff
@@ -368,7 +371,21 @@ class ToolOptions extends Component {
...
+              id="allowDangerousHtml"
+              name="allowDangerousHtml"
+              type="checkbox"
+              onChange={(e) => {
+                this.handleInputChange(e);
+              }}
+              checked={this.state.allowDangerousHtml}
+            />
+            &nbsp;
+            <label htmlFor="allowDangerousHtml">Tillåt HTML i infoclick</label>
```
sweco-seqtob, line 387: Thought we didnt allow html at all now, only markdown?

new-client/src/components/FeatureInfo/FeatureInfoContainer.js
```diff
@@ -4,8 +4,8 @@ import { withStyles } from "@material-ui/core/styles";
 import ArrowLeftIcon from "@material-ui/icons/ArrowLeft";
 import ArrowRightIcon from "@material-ui/icons/ArrowRight";
 import FeaturePropsParsing from "./FeaturePropsParsing";
-import Diagram from "../Diagram";
```
sweco-seqtob, line 7: We are not supporting diagram anymore?

new-client/src/components/Window.js
```diff
@@ -516,6 +516,7 @@ class Window extends React.PureComponent {
             {features && features.length > 0 ? (
               <FeatureInfoContainer
                 features={this.props.features}
+                options={this.props.options}
```
sweco-seqtob, line 519: Not necessary but we could destructure props instead

new-admin/src/views/tools/infoclick.jsx
jacobwod: There's support for both formats in ReactMarkdown. I'd like to shut it off, but it would meet a lot of resistance in the admins camp, so we give them an option to keep it on.

new-client/src/components/FeatureInfo/FeatureInfoContainer.js
jacobwod: Neither this nor the `Table` shortcode. I asked prior this change and no-one was using it.

new-client/src/components/Window.js
jacobwod: `{options}={{this.props}}`? Why?

sweco-seqtob: Sorry i meant all the props, but not necessary as i said

`const  {options, features, onDisplay, globalObserver}` = this.props;

PR #732, opened 2021-03-31
Fix/snackbar in mobile

### LINKED ISSUE #731: Snackbars covers window controls (On mobile)
**Describe the bug**
On mobile:
Sometimes the snackbars makes it very hard to maximize/minimize/close windows.
The buttons gets covered.

**To Reproduce**
Found issue in Navigation and Show Coordinates plugin etc.

**Expected behavior**
For Breakpoint XS and down:
I would suggest that the bottomCentered snacks gets approx 35px from bottom instead of 14px.
And that the topCentered snacks gets approx 35px from top instead of 14px.

**Screenshots**
notistack1
notistack2

jesade-vbg:
Proposed quick fix for #731 

I will probably look into this more in detail in the future, but I hope this quickfix will solve the immediate problem.

Results:
notistack3

Closes #731

### DIFF
--- a/new-client/src/components/App.js
+++ b/new-client/src/components/App.js
@@ -169,6 +169,16 @@ const styles = (theme) => {
     widgetItem: {
       width: "220px",
     },
+    snackbarContainerBottom: {
+      [theme.breakpoints.down("xs")]: {
+        bottom: "35px",
+      },
+    },
+    snackbarContainerTop: {
+      [theme.breakpoints.down("xs")]: {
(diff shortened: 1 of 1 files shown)


Hallbergs (approved):
Looks like it will get the job done!

jesade-vbg:
Related post-fix:
https://github.com/hajkmap/Hajk/commit/eef7fa265a4dbb0967439bde140715268f8d0c85
for reference.

PR #741, opened 2021-04-07
Feature/482 improved coordinates

hitomin21:
Modified coordinates tool to add a button for going to user location and fixed some bugs by restructuring the code. I created a new react component which shows a transformation. Any changes in the input fields are shared between the transformations using an observer so no dependency on html layout anymore and cleaner code. A button for removing all values inside the fields has been added and a button for moving the marker to the user's estimated location as well.

The tool looks like this now,
bild

Two new settings.
thousandSeparator: If checked in admin will format numbers 12345 -> 12 345
showFieldsOnStart: If checked in admin will put a marker at the center of the screen and fill in the coordinates for that location in the fields as an initial position.

thousandSeparator use react-format-numbers so I added it to package.json.

Admin has these new settings now (the two checkboxes in the middle). Default is disabled.
bild

My git is formatting Contributing.md when I commit to "fix it" but the contents should be the same if you view it in github.

Issue for the changes:
https://github.com/hajkmap/Hajk/issues/482


jacobwod (changes requested):
Nice work!

I merged the latest `develop` as there's been some changes since your PR yesterday, and it works fine!

### Small issue to fix
One small issue I ran into is that **when we just keep existing settings in admin, the `precision` key seems to be lacking**, which makes Coordinate to display everything with precision `0`. I'm not sure, but I guess that there must have been some other default previously, as my settings showed 3 decimals, if I recall correctly. **Perhaps you can change that, so we default to something more than 0 decimal points?** It's a minor issue but would make it easier to migrate without re-saving all config files.

Skärmavbild 2021-04-08 kl  11 28 22


### Bonus issue
It seems like we're using the `code` for each given projection in the loop that renders components. The value from `code` is used as the `key` inside the loop. It's not a problem, until someone tries to configure the same projection twice, if so, we run into this problem:
Skärmavbild 2021-04-08 kl  11 29 58

Of course, it can seem like a dumb thing to do as an admin (to configure the same projection twice), but we can never be sure what needs arise. This can be fixed in two places:
- The easy way: inside the loop in client. Don't use `code` for key, just use the `index` from `for/map/forEach`
- The harder way: in admin. Display a warning and don't allow saving if the same projection is setup twice.

Either way, it's a bonus issue, feel free to fix if possible but it's not necessary.

jacobwod (approved):
Nice work @hitomin21!

PR #746, opened 2021-04-08
Adding FME Server Download plugin

maan002:


jacobwod (approved):
I've tried it out with your demo from readme and can confirm that it **loads** the plugin and **runs** the app as before. I could not test the functionality _inside_ the plugin itself, but that'll be easier after we've merged it and I can deploy a build to my GIS colleagues. 

Please note: I am aware that this is a deviation from Hajk's standard (where we use to say that every stable plugin should be admin-configurable), but in this case the demand to try it out is high enough to set that rule aside.

PR #759, opened 2021-04-16
Fix/admin missing print settings

jesade-vbg:


jacobwod (approved):
I assume you've checked out this and tested locally, approving.

Issue #535, opened 2020-06-08
Edit plugin: possibility of locking WFS layers with AD groups is missing in Hajk 3


FatemeHosseini:
**Describe the function that is missing in Hajk3**
In Hajk2, the administrator can allow that a user (with AD) have access to the edit plugin and **to have access to a wfs layer (edit layer) in the plugin** but in Hajk3 can only allow user to have access to the plugin or not. 

![image](https://user-images.githubusercontent.com/36195301/84009964-4ceb2200-a974-11ea-9d79-0121186f99d6.png)

perosoderstrom:
Hi @jacobwod. I have made a test and implemented locking of edit layers with AD, in new-admin and new-backend. You can see the first test in branch ad-edit. Please notice this is just a first test, not ready for review yet. Now I saw that there is a new config.service.**v2**.js. I assume I shall implement the code there too? Comments?

jacobwod:
That's right. In case you wonder why it's separated and seemingly duplicated: I'm preparing a new version of the API that _will break backwards compatibility_ with the current backend API. As you see, the request to /config/:map in v2 will give you not only the map config but also layer definitions (of used layers) and some more objects. I guess that we want to keep both options alive for the time being, hence two files. 

I was thinking about inheritance, (so v2 would `extend` the v1 class), I'll see if I'll refactor. For now, it's two files. I highly recommend that you try out the new API, just set the `experimentalNewApi: true` in client's `appConfig.json`.

jacobwod:
Closed as of #762

PR #762, opened 2021-04-20
Edit plugin: possibility of locking WFST edit layers with AD

persoderstrom:


jacobwod:
Hi @perosoderstrom! This PR looks way better than before. But it's a big change that needs testing, so allow some time for us to test this from your feature branch (ad-for-edit-layers) _before_ we merge this into develop. OK?

persoderstrom:
Good to hear! Take your time to test!

jacobwod (approved):
I've done the following tests:
- NodeJS backend
- Latest code from develop (merged in to this branch prior my checks)
- Tested with `AD_LOOKUP_ACTIVE` both `true` and `false`
- Tested with `"experimentalNewApi"` both `true` and `false`

Everything seems to work out fine and behave as expected:
- With AD lookup _inactive_ all WFST services that are active in map will be shown
- With AD lookup _active_ only those services that user has access to are shown
- With the new consolidated loading (v2 of API) the WFST layers that are not needed are _not_ visible in the layers repository.

I assume you've developed towards the .NET backend (I've seen changes in those files). 

**I've noticed that keeping old data can lead to unexpected behaviour. Deactivating and activating the Edit plugin solves this.**

Good job!

persoderstrom:
Thanks @jacobwod and @Hallbergs for reviewing.
This PR should close #535.

Issue #765, opened 2021-04-22
Autocomplete in Search can't handle non-string properties


sweco-seqtob:
**Describe the bug**
When a feature has a non-string property the function encodeCommas ( fail because it tries to run string.replace (StringCommaCoder.js)
This appeared when a feature had an Number as the value for a property.

_**Would like some input on proposed solution, see screenshots**_

**To Reproduce**
1. Set up WFS service with a property not being a string
2. Make a search in Hajk
3. Console shows error in encodeCommas function

**Expected behavior**
The autocomplete in search should be able to handle non-string values

**Screenshots**
This is the current code that is not working
![image](https://user-images.githubusercontent.com/22886522/115673445-eb922e80-a34c-11eb-9cda-ef0b7a1b3c66.png)

Proposed solution
![image](https://user-images.githubusercontent.com/22886522/115673502-ff3d9500-a34c-11eb-984a-0e0032ec3575.png)

**Desktop (please complete the following information):**
 - OS: W10
 - Browser Chrome
 - Version 89.0.4389.128

PR #767, opened 2021-04-22
Handle non-string values in autocomplete

sweco-seqtob:
Proposed fix for #765


Hallbergs (approved):
Should get the job done!

jacobwod (approved):

PR #769, opened 2021-04-22
Vt/core changes 5

sweco-semtto:
Core changes from Västtrafik


Hallbergs:
@sweco-semtto We've merged your changes now. The plugin seemed to start but we could not test it properly. Make sure everything works and push changes if needed.

PR #770, opened 2021-04-23
Feature/761 update react markdown

jacobwod:
Let's try to get this into 3.6 rc2, as the older implementation seems to have some rough edges.


jesade-vbg (approved):
Great work, great improvement!

Hallbergs (approved):
Great work, as always

PR #774, opened 2021-04-26
Originalrepo/bug/backgroundswitcher

sweco-seqtob:
PR to fix the bug that made it impossible to switch between background layers if they've been defined with new backend.


jacobwod (approved):
Look great!

Issue #777, opened 2021-04-27
Make it possible to let autocomplete in search use wildcard at start


sweco-seqtob:
**Is your feature request related to a problem? Please describe.**

- Make it possible to let autocomplete in search use wildcard at start.

**Describe the solution you'd like**

- The autocomplete should have an option to enable wildcard before the actual lookup value.
- Add configuration for the search tool to enable wildcard at start for the autocomplete lookup.
- Default should still be to do the lookup for autocomplete without wildcard, to not break current functionality.

**Current state  for autocomplete**

**SearchString** : _Liseberg_
**Hits** : _Liseberg parking_
**No hit**: _Parking Liseberg_

**Propsed result**

**SearchString**: _Liseberg_
**Hits**: _Liseberg parking_
**Hits**: _Parking Liseberg_

**Propsed configuration**

![image](https://user-images.githubusercontent.com/22886522/116255406-7f5f6280-a772-11eb-8144-3132b4e0ed49.png)

sweco-seqtob:
This is functionality requested and i can implement it as long as no one is opposed the idea.
Default is still the same as today and this is only optional

Hallbergs:
I like it

PR #779, opened 2021-04-28
Originalrepo/feature/wildcard at start for autocomplete

sweco-seqtob:
Added the possibility to add wildcard at start for autocomplete.
Default is still `wildcardAtStart = false` for autocomplete if not set in admin/config

Solves #777


jacobwod (approved):
Lovely!

jacobwod:
Ooops… now I see that this went all the way to `master`. What do you say @Hallbergs, ok or should we revert and push toward `develop`?

Hallbergs:
Since it defaults to false (and the users who doesn't update their config won't notice the difference) i think we can include it in rc2.

PR #783, opened 2021-04-29
This will remove the <br /><br /> tags from Infobox

FredrikEkstrom79:
This will remove the text "<br /><br />" from layers infobox in layermenu that in some cases will override the infoclick to not show any information for selected feature.


jacobwod (approved):
Well done! I'll take a closer look and probably merge to master as a bug fix!

Hallbergs:
It might be that im a bit tired, but shouldn't it be `input4 = $(`<textarea id="${id7}" type="text" />`)` and not `input4 = $(`<textarea id="${id7}" type="text" /></textarea>`)`?

jacobwod:
You mean the shorthand closing HTML tag as the sole difference? Since the text node inside the tag is empty, they're both equal, so sure, you're right!

jacobwod (approved):

PR #788, opened 2021-05-06
Feature/756 visualize layer minmax zoom hidden

jacobwod:
I like what @jesade-vbg has done here, I think we're ready for a PR.


Hallbergs (approved):
LGTM

Issue #790, opened 2021-05-07
Vertical scroll is not working on Android.


jesade-vbg:
**Describe the bug**
As the title says... vertical scroll is not working on Android.
This makes it impossible to access all layers in layerswitcher if you have more than ~20 layers.

I am currently working on a fix that will be pushed to a new branch.

PR #791, opened 2021-05-07
Fix 790. Android scroll issue

jesade-vbg:
#790 

I've been testing on an iPhone SE (2020) and on two diffrent Android phones with good results.

You are welcome to test aswell.

Can we keep this fix branch after merge?


jacobwod (approved):
Did some quick test on iOS 14 and it seems to work as expected. I assume that the Android bug is solved too. Closes #790.

PR #798, opened 2021-05-12
Fixes #797

### LINKED ISSUE #797: Infoclick does not update value
The infoclick box does not update the values in the latest version of develop. You have to close the info window and click again to get a refreshed value.

It can be reproduced using the default layers.json and map_1.json with the layer Riksintresse Kulturmiljövård.

Hallbergs:
The reason that we couldn't infoClick arcGis features was that the id-key on the parsed features was missing. This should take care of the problem. Only fixed for geoJson-format so far.


jacobwod (approved):
Wonderful!

PR #800, opened 2021-05-18
Demo site update

jesade-vbg:
Contains some changes to the configs and a quickfix for the argis hidpi issue.
Hajk-samverkan layer added and visible at start.

Should be ok to merge to master as far as I can see.


Hallbergs (changes requested):
42 commits for demo site update? I think not :)

Issue #802, opened 2021-05-18
Breadcrumbs uses hard-coded colors


Hallbergs:
Since the breadcrumbs are using hard-coded colors, they look messed up when using dark mode.

PR #803, opened 2021-05-18
Fix/802 hard coded colors bc

Hallbergs:
Fix for #802 

Not sure if the non-rounded edges of the breadcrumb is nice? What do you think? Everything else is not rounded so maybe this fits better with the overall look and feel of Hajk? Hmm...

Would be nice if someone with the informative plugin configured could check if the links etc render nicely.


jacobwod:
@jesade-vbg Do you have a map config with the old Informative plugin active? If so, can you test if the Breadcrumbs still work in such configuration?

Otherwise, let's merge as we're phasing out the Informative plugin anyway, @Hallbergs?

jesade-vbg:
@jacobwod Sorry I have never had that active or in buildconfig :/

PR #804, opened 2021-05-19
Hotfix: Geometry can be null for raster causing fatal crash on mobile…

jesade-vbg:
A tiny null check.


Hallbergs (changes requested):
The tests failed. I think a space is needed: if(geom) => if (geom)

jesade-vbg:
> The tests failed. I think a space is needed: if(geom) => if (geom)

Prettier had crashed, failing to format my code :)

Hallbergs (approved):

PR #829, opened 2021-06-22
Added WFS-T support for QGIS Server

### LINKED ISSUE #825: Add support for WFS-T from QGIS Server
Hajk has drifted towards a GeoServer-centric approach recently, and great as GeoServer is, it's always good to give the users more options. 

Adding WMS layers from QGIS is supported out of the box (though the admin experience could be improved). WFS Transactions however, require some code changes:
- Most notably, QGIS Server follows the WFS specification strictly and lacks support for `outputFormat: "application/json"` for `describeFeatureType`. We must handle the XML response in admin in order to be able to add layers.
- We must remove the `geometry` field somehow. The current approach is no good as QGIS Server will return the specific `type` of filed (e.g. `gml:MultiLayerSurface`), and not just `Geometry` as GeoServer seems to do. At the same time, we must keep the current support for the GeoServer style.
- On successful write, GeoServer always returned all types (`totalInserted`, `totalDeleted` and `totalUpdated`), while QGIS Server only returns the changed type. We can't expect all the properties to exist, so `client` needs some minor fixes too.

Overall, quite an addition, but I'll attempt to close this before the weekend.

jacobwod:
Closes #825, see issue description for more details.

### DIFF
--- a/new-admin/src/models/edit.js
+++ b/new-admin/src/models/edit.js
@@ -22,7 +22,10 @@
 
 import { Model } from "backbone";
 import $ from "jquery";
+import X2JS from "x2js";
 import { prepareProxyUrl } from "../utils/ProxyHelper";
+
+const x2js = new X2JS({ attributePrefix: "" });
 
 var edit = Model.extend({
   defaults: {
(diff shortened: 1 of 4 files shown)


Hallbergs (approved):
LGTM (im no WFS spec-expert though :) )

Issue #807, opened 2021-05-21
Invertera not working in coordinate tool


hitomin21:
It seems like the invertera setting for a transform is ignored in the coordinate tool. The order of the coordinates are not swapped regardless of the value specified in admin.

FredrikEkstrom79:
Is this realy inverted? seems to be only the order the values are presented to the user that is changing.
![image](https://user-images.githubusercontent.com/50202826/124099097-771af080-da5d-11eb-8529-9ac30747b6d9.png)
![image](https://user-images.githubusercontent.com/50202826/124099165-869a3980-da5d-11eb-8a23-bc144b9e2925.png)

jacobwod:
Good find! I agree: just shifting the display order is not the same as inverting axis.

PR #835, opened 2021-06-22
Feature/807 coord invert

hitomin21:
Fixes a bug for inversing coordinates in the coordinates tool.

The bug was that this.props.inverseAxis was not set when the CoordinatesTransformRow was created.


jacobwod (approved):

jacobwod (approved):

jacobwod:
Reverted, see #807 for reason.

PR #839, opened 2021-07-05
Fixed inverse axis

### LINKED ISSUE #807: Invertera not working in coordinate tool

FredrikEkstrom79:
Have put back the changes that was made in #835 exept for the labels changing place, now the values for N/E will change but labels for N/E will remain in the same order as what was discussed in #807.
image
image

Closes #807 #835 #837


matself:
Etikett stavas￼ så här.

jacobwod (approved):

Issue #845, opened 2021-07-26
Selected Buttons become black when selected in edit tool


hitomin21:
The buttons (Punkt, Flytta, etc.) color changes to black once they are selected in the edit tool, see below.

![bild](https://user-images.githubusercontent.com/22234673/127013652-a5c3b01b-4cc5-4417-93e5-5b7c1bf6723e.png)

I will change to a lighter color since it is difficult to read the text on the selected button.

hitomin21:
Using https://github.com/hajkmap/Hajk/tree/fix/845-change-selected-color-edit for the fix

Issue #882, opened 2021-10-19
Search plugin not working for QGIS Server


WCarlsson:
# Edit
This became quite big as we realised that what we _really_ want to do is rewrite how search results are handled inside Hajk. Currently the entire search result flow is built around – and expects – GeoJSON objects. A better, and more format independent solution, would be to accept any standard geo format as response and convert that to a common ground (i.e. OpenLayer's `Feature`s) before going further down the flow.

So what we'll be doing now is exactly that. Search will be rewritten to work around OpenLayer's features, and it will accept both GML and JSON responses. 

As it is a quite substantial rewrite, I'm creating a separate feature branch from here: https://github.com/hajkmap/Hajk/tree/feature/882-search-uses-ol-features. 

# Original issue
**Describe the bug**
When using the search plugin with a QGIS Server layer there is no result shown in the search bar. The response from the server is in gml-format but it seems Hajk expects json-format.
We have tried to add outputFormat=application/vnd.geo+json to the request URL but the response is still in gml-format.

**To Reproduce**

1. Set up a search layer from QGIS Server in Hajk admin and activate it in a map
2. Search for a value that should give a hit
3. See that no value populates the search bar
4. If you look at the response you can see a hit in gml-format

**Expected behavior**
The search bar should be populated with the expected values.

**Desktop (please complete the following information):**

- OS: Windows 10
- Browser: Any

Hallbergs:
@matself  It would be great with some open data from QGIS server so that we could run some more tests!

jacobwod:
Just some more work to do in order to get the spatial search going. QGIS Server seems to be very formal in which request bodies are allowed: if the polygon sent in the request has duplicate points, QGIS won't give any response at all. See below, in the first screen shot **the first and last points are duplicates** and there is no response. As soon as I remove the duplicate, I get the correct response. 

### Duplicate
![Skärmavbild 2021-11-01 kl  11 03 09](https://user-images.githubusercontent.com/110222/139655355-1dbc683b-e092-4524-9f1f-e1b4db718f44.png)

### No duplicate
![Skärmavbild 2021-11-01 kl  11 03 29](https://user-images.githubusercontent.com/110222/139655367-cd8e72ef-f878-4768-bc29-7d79da8167ef.png)

jacobwod:
I'm ready to merge it to `develop` but would like to see some test before. Can some of you please checkout `feature/882-search-uses-ol-features` branch and test locally? @WCarlsson

[...]

Issue #853, opened 2021-08-26
Edit: Support for more geometry types


hitomin21:
I would like us to support at least the following geometry types for the edit tool:

Point, LineString, Polygon, MultiPoint, MultiLineString, MultiPolygon

Being able to enter multiple geometries in one edit would be beneficial in many situations.

Issue #847, opened 2021-07-29
Edit: Support for common datatypes


hitomin21:
I would like us to support at least the following for the values written into the edit tool:

int, number, string, date, date-time

PR #884, opened 2021-10-21
Feature/847 support for common datatypes

hitomin21:
FYI: The reason that I created one, instead of several, pull request to solve the issues below is that most of the changes (which I originally made in different branches) were in the same files and added new functionality which I wanted to test together (for example testing that the alias and descriptions on the new data types as well as re-submission with all new data types). Instead of creating individual pull requests and having to merge everything again I have therefore submitted this branch which is merged and bugs, caused by the new features interacting with each other, fixed.

Adds support for multi-(line, point and polygon). This does not allow a user to insert multiple geometries in the edit tool but allows us to use the edit tool where the backend expects a multi* object as input.

Adds support for showing non-editable fields to the user and also submits these (and hidden fields) with their default values.

Adds support for more data types such as date.

Adds support for giving aliases to fields so the database column names are not used so we can give better field names that are shown to the user. Also adds a description for each field.

Adds an option for a user to continue to edit a submission in case the submit fails.

Fixed bug where entering something in instructions for edit tool made admin crash

Closes issues:
https://github.com/hajkmap/Hajk/issues/845 (black button)
https://github.com/hajkmap/Hajk/issues/846 (alias)
https://github.com/hajkmap/Hajk/issues/847 (Support for more datatypes)
https://github.com/hajkmap/Hajk/issues/853 (Support for Multi-* types)


jacobwod:
@hitomin21 I've done some serious changes to how search works (to fix #882). Make sure to merge `develop` in this branch as soon as possible.

jacobwod (approved):
I've done two fixes to avoid crashes and merged develop. I'm not sure that we've got editing to work properly with layers from QGIS Server yet - more tests are needed and I think it's time for a RC so this can get more proper testing. Merging.

new-client/src/plugins/Edit/EditModel.js
```diff
@@ -353,30 +353,28 @@ class EditModel {
...
       srsname: source.projection,
-    });
+    };
+
+    // Create a new URLSearchParams object from the merged object…
+    const searchParams = new URLSearchParams(mergedSearchParams);
+    // …and update our URL's search string with the new value
+    url.search = searchParams.toString();
+
     hfetch(url)
```
jesade-vbg, line 378: Url needs to be a string, now hfetch gets an object. Edit plugin fails.

PR #911, opened 2021-11-18
Feature/855 fir plugin

jesade-vbg:
@maan002 will make a separate PR for backend related changes (later this day).


Hallbergs:
Just out of curiosity, this plugin will only work if you use the .NET-backend?

jesade-vbg:
I would probably work with nodejs backend, except for the exports, EDP etc, and the AD-filtering which has specific code in .Net.

Hallbergs:
Alrighty!

Hallbergs (approved):
Just made a quick overview. LGTM.

PR #912, opened 2021-11-18
Feature/908 enhance print plugin 3

jesade-vbg:


Hallbergs:
I added a small fix regarding the scale calculation which seemed to do the calculations on a static scale that is set when the printView is initialized.

Hallbergs:
Also, i am not sure that we have to store the original view-information (such as the resolution and center) now when we're using the print-view instead of the original one. The same applies for `constrainOnlyCenter`, it is set to `this.mapConfig.constrainOnlyCenter` when initializing the print-view, but then we make sure to set it to false while printing. Maybe we could just set it to false right away. Thoughts?

I will investigate this further when i get a chance. 

Everything seems to work after the scale fix though! :)

jesade-vbg:
Hi, great find and nice comments :) I forgot to comment. Let me know next time so I can fix it :) yes I know we can clean up the code, the approach of using a new view eliminates the need to keep track of those params you mention.

jesade-vbg:
I actually havn't noticed anything wrong with the prints with the bug you mention. I've printed PDF, printed in papper and measured many times.

Hallbergs:
The preview that shows what is about to be printed would be static without the fix. The preview should be set to a fitting size which depends on the users current zoom. I could be wrong though.

Hallbergs (approved):

Issue #344, opened 2019-11-26
Add an option to search only in visible layers


victoraxbom:
When searching with the search plugin, there should be an option for the user to search only in those layers that is activated in the map. 

The activation checkbox could for example be situated under the "more options" in the search bar.

PR #914, opened 2021-11-19
Feature/search in visible layers 

perosoderstrom:
A solution to issue #344


jacobwod:
I've made a minor tweak and have also a couple of propsals in comment. Let me know if you want to talk about this!

perosoderstrom:
Good! I am out of "home office" a couple of days but will have a look when I'm back.

perosoderstrom:
Hi @jacobwod. I've added a filter to get search sources that are visible. Can you have a look?

jacobwod:
Sure, I'm looking into this. There seem to be cases that need to be handled, else they lead to something like this:
Skärmavbild 2021-12-02 kl  10 19 03

I'll see what causes this.

jacobwod:
Unfortunately there are some issues with this solution. 

### 1
If only one sub layer is selected, like this:
Skärmavbild 2021-12-02 kl  10 27 54

… the `getVisibleLayers()` grabs the entire layer, including all sub layers, and will compare against them:
Skärmavbild 2021-12-02 kl  10 26 33

### 2
I don't think this can work, as `l.subLayers` is an Array of strings, and that is not a valid key in the `layersInfo` object:
Skärmavbild 2021-12-02 kl  10 31 36

perosoderstrom:
Ok. Our test layers contains only one subLayer and that works. Do you have any hints to solve this, @jacobwod ?

jacobwod:
Well, I think we could do something like this. 

You have the search sources, that are combined WFS and converted WMS services. They all seem to have a unique ID property:

image

So comparing to this ID would be fine, I guess. 

What you want to do is **change what `visibleLayers` variable holds**. Currently it holds too much information (we only need the ID of each sublayer).

So I'd try changing the `getVisibleLayers()` method to return an Array of strings that are IDs of visible sublayers.

Currently it's an Array of objets (2 objects, in the screenshot below), and even then it's not correct (as it contains all `subLayers` even though only one is active):
Skärmavbild 2021-12-02 kl  10 52 39

Instead of that, you want something like:
```javascript
["sublayerXfromLayerY", "sublayerAfromLayerZ"]
```

If you've got this part correctly and `getVisibleLayers` returns a flat Array, the filtering will become much easier.

jacobwod:
> Is there a way to check if a single sublayer in a group is visible (besides traversing the layerswitcher)? 

Currently we don't have any trivial way of doing this check (although we should write a helper and put in `utils`). 

Each `Source` has a `LAYERS` parameter, that controls which sublayers are active. We set them using the `updateParams()` method on the `Source`:
https://github.com/hajkmap/Hajk/blob/f899720b3a4cb354cea6bbdeefa8fbb5e44175ef/new-client/src/plugins/LayerSwitcher/components/LayerGroupItem.js#L436-L447

**To see which sublayers are active, you must look at the `LAYERS` parameter** that accessible on the `Source`, (e.g. `mapLayer.getSource().getParams()`, than extract `LAYERS` from there). `LAYERS` will (probably) be a comma-separated string, so you must do a `.split()` to make it an Array. The resulting Array can then be compared against the results from `getVisibleLayers()`, that now also is an Array of strings.

jacobwod:
@perosoderstrom, I haven't checked out the code this time but I took a quick look. You seem to be using the "private" OpenLayers properties (those ending with underscore `_`). Please don't do that. To access the source, do `layer.getSource()`. To access a property in values, to `layer.getSource().get('someProperty')` etc.

Regarding `params`, there's even a `getParams()` method on the `Source`, please use it.

jacobwod (approved):

perosoderstrom:
Thanks, @jacobwod !

jacobwod:
Good job @perosoderstrom! We've just made some minor fixes, such as adding a notification so the user is aware when the limitation is active. Works like a charm. 👍 
Skärmavbild 2021-12-10 kl  11 28 16

PR #919, opened 2021-11-22
Fixes #916

### LINKED ISSUE #916: Ensure that XLSX export in Search prints all the columns
Not all rows may have data in all columns. Just grabbing the key names from the row with most values might not be enough:

```
example
```

> We might still miss keys... I'll investigate further. @jacobwod
_Originally posted by @Hallbergs in https://github.com/hajkmap/Hajk/discussions/907#discussioncomment-1669144_

Hallbergs:
This should get 916 fixed. What do you think @jacobwod?


jacobwod (approved):
Works like charm!

PR #930, opened 2021-11-25
GeoServer Proxy from Hajk2 migrated to Visual Studio 2019.

### LINKED ISSUE #929: Add GeoServer proxy from Hajk2
The proxy "GeoServerProxy" found in branch hajk2.x in folder proxy/GeoServerProxy is missing in Hajk3.

The proxy should be migrated to Visual Studio 2019 before added to Hajk3.

maan002:
Closes #929


Hallbergs (commented):
Are you/do you know if anyone is using this proxy? If i understand it correctly it's only purpose is to catch certain http-requests and pass them to some pre-defined path on the local server? Aren't there better solutions in IIS for a simple task like this? (ARR, url-rewrite?)

If you are using it then of course, we'll add it, but otherwise i don't really see the point. Please enlighten me if i've failed to understand the purpose of it.

jacobwod:
I've checked out this code as well and it looks to me, just as @Hallbergs points out, that there are better solutions for proxying certain requests than an entire .NET solution.

maan002:
This proxy is used to access protected layers in GeoServer.
In order to support this the proxy sets a header with the users name.
ARR does not support this since it is invoked before IIS authenticating process.

Also, the way this proxy does this makes it possible for QGIS to access the protected WMS layers as well.

Hallbergs:
Oh, ok. I only read the index.html and didn't really get the purpose. It's exactly as you say, the authenticated user cannot be accessed in ARR (which is a bit of a shame tbh). 

I guess the proxy serves a function then, and we should merge it. Could you add some kind of documentation so that it would be easy for someone else to use it? :)

Hallbergs:
The readme looks good. Ready for merge?

maan002:
Yes, please merge!

Hallbergs (approved):
(y)

PR #937, opened 2021-11-30
Dop develop

sweco-semtto:
A PR for document handler. Added new tags and updated the print function.


jacobwod:
Correct me if I'm wrong but this looks like a bulk PR? 

My understanding was that in the begging of the project we decided that you'd push smaller changes more frequently?

sweco-semtto:
You are right that this PR is a big one and you are also right that our intention was different. Unfortunately, it was not a feasible course of action that we could apply.

Hallbergs:
Just a quick update. Will take a look at this tomorrow!

sweco-semtto:
Great, please let me know if I can help.

Hallbergs (changes requested):
I was about to start reviewing, but upon starting the Admin UI i was prompted with several warnings, see below:

image

Please take care of these, then i'll have a look at the code.

jacobwod:
The printout itself works and is looking good! 

The Admin UI, however, refuses to load since you haven't implemented the new endpoints in the NodeJS Backend.
Skärmavbild 2021-12-10 kl  10 17 25

What's your opinion on this? **Was handling of video and audio tags only ment to work in the .NET backend?** If so, why did you add the new endpoints to the `new-admin/public/config.docker.json`:
```json
    "list_videos": "/api/v1/mapconfig/listvideo",
    "list_audio": "/api/v1/mapconfig/listaudio"
```

This will not work and will lead to the Docker container crashing, as it's NodeJS backend-only.

sweco-secoma:
We have fixed the lint warning messages in admin and added routes for video and audio in the new backend. There is no file type checking for the audio/video routes (same as the existing image route).

backend/mapservice/Controllers/ConfigController.cs
```diff
@@ -655,5 +665,59 @@ public string ListImage()
...
+
+            IEnumerable<string> files = Directory.GetFiles(folder);
+            List<string> fileList = new List<string>();
+            foreach (string file in files)
+            {
+                string fileName = String.Empty;
+                if (Path.GetExtension(file).ToLower() == ".mp4" || Path.GetExtension(file).ToLower() == ".mov" || Path.GetExtension(file).ToLower() == ".ogg")
+                {
+                    fileName = Path.GetFileName(file);
+                    if (fileName != "layers")
```
jacobwod, line 687: Why?

sweco-semtto: To diffrate pictures from videos etc. that are in the same folder.

Hallbergs: I think @jacobwod is referring to the `if (fileName != "layers")` part. The configuration files are not in the same folder as the upload-data, right?

sweco-semtto: Correct, the configuration files are not in the Upload folder, only data files such as images and videos. I guess this line is from an previous iteration of DOP.

sweco-semtto: I will fix this issue and update the PR with that fix.

sweco-semtto: Done.

Hallbergs (approved):

PR #948, opened 2021-12-06
Feature/944 min max zoom disable warning snackbars

jesade-vbg:


jesade-vbg:
Should be pretty straight forward.

Hallbergs (approved):

Issue #878, opened 2021-10-14
Symbols for vector layers not working


perosoderstrom:
**Describe the bug**
Vector symbols with offset can be used when you want a pointer symbol 
to point exactly at the spot.

In Hajk 3.2 it was possible to set a symbol for points in vector layers. 
You could also set an X/Y Offset for the symbol. 

When uppgrading to Hajk 3.7 the symbols does not show on the map.

**To Reproduce**

1. In admin add a vector layer "Vektorlager" with points and data format WFS.
2. Set an "Ikon": i.e. https://karta.kungsbacka.se/publik/assets/map_icons/forskolor.svg
3. Notice that there are no fields in admin for X/Y Offset of the symbol.
4. Add the vector layer to a map.
5. When you open the map and turn the layer visible, no symbols are shown.

**Expected behavior**
Symbols whith a given offset should appear on the map.

**Screenshots**
1. Hajk 3.2 map with vector symbols
![Hajk 3 2 karta med vektorsymboler](https://user-images.githubusercontent.com/51197693/137288015-e3127b23-6e91-44ca-bf01-0ba52fa672e8.png)

2. Hajk 3.2 Admin 
![Hajk 3 2 admin](https://user-images.githubusercontent.com/51197693/137288098-275f1f32-d8d1-4f90-abc9-a31c30ec2e0f.png)

3. Hajk 3.7 Admin
![Hajk 3 7 admin](https://user-images.githubusercontent.com/51197693/137288140-d5e60e92-cb53-4c25-8687-e9f3dc833c2b.png)

**Desktop (please complete the following information):**

- OS: Windows 10
- Browser: any

**Additional comments**
This might have been changed in "Feature/592 snap interaction #598" commit.

jacobwod:
Hi @perosoderstrom, thank you for such a detailed description! It surely helps me to track down this bug, I'll take a look as soon as I have a chance. Great find! 👍 😄

jacobwod:
> Yes, a default marker.
> ![Hajk 3 7 vektorlager med default symboler](https://user-images.githubusercontent.com/51197693/139062210-1d0c9383-bc50-4832-a162-ec37f97d6d5c.png)

OK, so you do get the default marker. Have you tried specifying the marker in the URL too? Try with absolute URL, make sure that you don't get 404 in the Network tab too. I'm pretty sure you will see your marker, it's possible to set that in Admin UI. 

The offsets however are missing. It that what you want to add?

[...]

PR #953, opened 2021-12-13
Feature/additional vector layer styling (#878)

perosoderstrom:
Restores previous vector layer styling.
See issue #878


perosoderstrom:
@jacobwod , @Hallbergs. This is blocking updating Kungsbackas public map. If possible, Kungsbacka would like to have this fix in 3.8 version.

Hallbergs:
> @jacobwod , @Hallbergs. This is blocking updating Kungsbackas public map. If possible, Kungsbacka would like to have this fix in 3.8 version.

Me, Jacob, and Per discussed this in a meeting yesterday. Some kind of solution will probably be presented in 3.8.

perosoderstrom:
Let me know if there is anything I can fix.

new-admin/src/views/layerforms/vectorlayerform.jsx
```diff
@@ -133,6 +158,10 @@ class VectorLayerForm extends React.Component {
       return new Date().getTime().toString();
     }
 
+    function rgba_to_string(c) {
```
jesade-vbg, line 161: Please use camelCase for consistency. rgbaToString

new-admin/src/views/layerforms/vectorlayerform.jsx
```diff
@@ -144,6 +173,8 @@ class VectorLayerForm extends React.Component {
     if (fieldName === "date") value = create_date();
```
jesade-vbg, line 173: Please use camelCase for consistency. createDate

perosoderstrom: Hi @jesade-vbg, 
Thanks for reviewing!
"create_date" is used in other *layerform.jsx and in search and edit.jsx so i suggest we leave it as is, for consinstency ;-).
I changed "rgba_to_string" to "rgbaToString" as you suggested in previous comment. Pushed it to github.

jesade-vbg: Great! Sorry I was'nt paying attention to where create_date was placed. Admin is a diffrent story.

jacobwod:
I'll check it out with our current SLD styling locally, to ensure everything works out fine with current solutions too. (Hopefully later today.) 

If so, it'll be ready to merge.

jacobwod:
The solution crashes when no values are specified for the newly added keys. In my case, e.g. the `icon` is `undefined`:
Skärmavbild 2022-01-20 kl  09 16 51

Which leads to this:
Skärmavbild 2022-01-20 kl  09 14 51

---
You should ensure that _all_ needed properties are set in config, before you try to create the custom style `this.style = createStyle.apply(this);`.

In addition, I don't see the point of having `createStyle` outside and doing this cumbersome `.apply()` call. Please place it as a usual class method.

jacobwod:
Looks like a good commit @perosoderstrom! I'll take a closer look tomorrow.

jacobwod:
Hi @perosoderstrom.

So the styling is correct if I supply SLD or point out an SLD file in Admin UI.

If I however remove them, I don't get the default OL-styling (the blue polygons), but rather something like this:
Skärmavbild 2022-01-25 kl  10 20 56

That's the case with _empty_ settings (or as empty as they can be - I can't remove some of these…). 
Skärmavbild 2022-01-25 kl  10 22 38

jacobwod (changes requested):
Apart from the behavior above, I've commited some changes (that don't fix the problem in my previous comment).

You can see the changes in VectorLayer.js:

- Since createStyle is now a class method, there's no need to use .call() or .apply(). Changed how it's invoked.
- Removed getStyle() as it was only called once, in the constructor. We can instead just use the value that resists in this.style.
- Even better: there's no need to save the value in this.style, as it is only used in once place (again, the constructor). Hence, line 115 now is simplified.
- As createStyle() is invoked only once and _without_ parameters, there is no need to have the parameters in method definition (see line 306). Removing them would lead to further simplification inside createStyle.
- What's even more interesting is that we have a lot of unused code in createStyle. E.g. everything that has to do with 'feature' is never invoked, as we never provide any 'feature' to the method.
- The entire comment with generateLegend() should be removed.

perosoderstrom:
Hi @jacobwod, 
A question about vector colors.
`lineColor: "rgba(0, 0, 0, 0.5)" `was used as default for vector layers prior to the SLD styling was implemented. It was defined in `new-admin/src/views/layermanager.jsx` and `new-admin/src/views/layerforms/vectorlayerform.jsx`. 
Shall we change default Hajk vector style to default OpenLayers style?

jacobwod:
> Shall we change default Hajk vector style to default OpenLayers style?

Hi @perosoderstrom. I would prefer the following chain of events: 
```
const style = custom config from Admin UI || custom SLD from Admin UI  || no style at all (i.e. fallback to OpenLayers default)
```

perosoderstrom:
I'm not quite sure what you mean with "custom config from Admin UI". 

In current PR it is implemented as
`const style = custom SLD from Admin UI || custom OpenLayers style from UI || no style, fall back to Hajk default`

But instead of falling back to previous Hajk default it will use OpenLayers default. Will that be OK?

jacobwod:
> But instead of falling back to previous Hajk default it will use OpenLayers default. Will that be OK?

I think we talk about different "previous" versions here. To me, "previous" is the state it was in prior this PR. (While I think that you refer to the previous implementation of custom styling, that has been removed when I implemented styling with SLD, right?) 

If so: yes, it's fine to just fall back to OpenLayers if no custom nor SLD styling is set.

jacobwod:
@jesade-vbg What's your opinion on this:
https://github.com/hajkmap/Hajk/blob/0ab97c590ed8966336b0088441bf735ec0b2cd3c/new-client/src/models/layers/VectorLayer.js#L18-L20

Is it really needed when `hfetch` is used?
https://github.com/hajkmap/Hajk/blob/0ab97c590ed8966336b0088441bf735ec0b2cd3c/new-client/src/models/layers/VectorLayer.js#L430

jesade-vbg:
same-origin is the default behavior of hfetch, so the three lines can be removed.

jacobwod:
I took a look at the code and corrected one minor thing (see previous comment). 

**Unfortunately, there's something going on in this branch that leads to total freeze upon app load.** Please the video below where I compare `develop` (which has no lag) with this branch. The app just freezes for about 10 seconds:

https://user-images.githubusercontent.com/110222/156149120-df540aab-1c4c-449e-b9aa-663436dbb1c5.mov

I'd be great if someone could confirm, ping @Hallbergs @jesade-vbg. *I have two WFS layers in this map config, both styled by SLD.* When I do the same test using a map config _without_ WFS layers, there is no freeze. So it's clearly related to `VectorLayer`.

jacobwod:
> It is probably these settings that are missing in map config. It does not fallback to correct values if not specified in config.
Thanks for pointing it out - it reminded my to merge latest `develop` into this branch.

Anyway, the problem persists. And I'm on the same map config as you can see in the video, only switching branches. The settings you refer are in `AppModel.js`, and that file is unchanged between this branch and `develop`.

perosoderstrom:
Just tested with a fresh clone. 
I added one vector layer and I can't see any freeze, without styling, with SLD styling or with OpenLayers styling.

Hallbergs:
We've located the issue. If the vector layer added contains alot of features, we block the thread when we try to parse them. See [this line](https://github.com/hajkmap/Hajk/blob/be48b886f8feb8c1022eef914f5703c8a2c2e07c/new-client/src/models/layers/VectorLayer.js#L429).

We could get around this by doing this in a worker, but we're not sure thats a valid solution since we're still fetching all the data just to create a legend graphic (we're talking several MBs of unneccesary data). Hmm...

jacobwod:
The problem is with how we generate the legend. Unfortunately, it results in two separate problems:
A. _A lot_ of data is loaded on app load (depending on the size of dataset of course, but potentially hundreds of thousands of objects)
B. The legend is generated even if separate SLD styling is supplied, which leads to discrepancy between what the features look like in map vs what the legend says they look like.

### A. Everything is loaded 
To generate the legend, the code does basically this:
1. gets all features
2. parses them
3. checks what we've got (points? lines? polygons?)
4. generates a symbol for each of the types found in step 3

This algorithm's weakness lies in the first point: the entire dataset is retrived from server. In your case, @perosoderstrom, where the dataset is small, you don't see the delay. In our case, where we get all buildings, **step 1 leads to 13 Mb being downloaded, while step 2 has to parse 103605 features**. That's what causes the freeze.

### B. Generated legend differs from SLD styling
Since the algorithm for legend generation is invoked even when SLD styling is used, we can get "funny" results, like those in the screenshot below:

bild

Hallbergs:
At a second thought, `DescribeFeatureType` probably won't work since the type returned is `Geometry` in many cases (which tells us nothing).

jacobwod:
> (Note: In `generateLegend `only the first feature is used to get featureType =` features[0].getGeometry().getType();)`

Well, in that case the algoritm isn't particularly clever. A simple `limit=1` to the WFS request would solve the risk of huge data loads. Still though, the generated legend _can_ be inaccurate as it only cares about the first element in the dataset. I feel this confuses the user and leads to a worse product than without this function.

**I propose that everything related to the `generateLegend` functionality is removed before the PR is merged.**  In order to ensure a proper legend, admins should generated a static image and point it (it's the `icon` property if I recall correctly).

jacobwod (approved):
Just pushed a change where legend generation is removed. This should be a good compromise for everyone.

PR #980, opened 2022-01-13
Feature/964 preset links changes

jesade-vbg:
Nothing weird. Pretty straight forward. Backward compatibility tested and works as far as I can see.


Hallbergs (approved):
LGTM

PR #981, opened 2022-01-13
Feature/856 kir plugin

### LINKED ISSUE #856: Port Hajk2 KIR plugin to Hajk 3
The KIR plugin in Hajk2 needs to be ported to Hajk3
(Edit: This will not be a port, it will be a complete rewrite with similar functionality.)

I will be created with similar functionality as in Hajk2. Varbergs original solution.
The design will of course match Hajk3 and use MaterialUI components.

Screenshot from Hajk2
KIR

jesade-vbg:
closes #856


new-client/src/buildConfig.json
```diff
@@ -36,9 +36,11 @@
     "Edit",
     "Export",
     "FmeServer",
+	"Fir",
```
jacobwod, line 39: Weird indentation? 😉

new-client/src/plugins/Kir/KirLayerController.js
```diff
@@ -0,0 +1,304 @@
...
+      bufferFeature.set("kir_type", "buffer");
+
+      _bufferFeatures.push(bufferFeature);
+    });
+
+    const targetSource = this.getLayer("buffer").getSource();
+    targetSource.addFeatures(_bufferFeatures);
+  };
+
+  _getZoomOptions = () => {
```
jacobwod, line 263: I like the separation between public and private methods. But rather then this custom "underscore" method, I'd prefer to use the correct [ECMAScript Private Class Fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields). (Basically, switch the `_` to the `#`.)

I comment only here but the comment is valid for all occurrences of `_foobar`.

new-client/src/plugins/Kir/KirSearchResultItemView.js
```diff
@@ -0,0 +1,104 @@
...
+        <div
+          className={classes.root}
+          dangerouslySetInnerHTML={{ __html: this.getHtml() }}
+        ></div>
+      </>
+    );
+  }
+}
+
+const styles = (theme) => ({
```
jacobwod, line 74: Best practice says that styling should be placed in the beginning of the document. 👍 👍 

(This comment is valid in your other `View`s as well.)

new-client/src/plugins/Kir/KirSearchView.js
```diff
@@ -0,0 +1,399 @@
...
+              </Button>
+            </div>
+          </AccordionDetails>
+        </Accordion>
+      </>
+    );
+  }
+}
+
+const styles = (theme) => ({
```
jacobwod, line 305: Not much of a comment, rather a thought: this amount of custom styling will make it increasingly difficult for you to migrate to MUI5 styling (if not difficult than at least require more rewriting). I'd try not to style as much and go with the default looks wherever possible. The defaults are usually very sane.

jacobwod (commented):
Nice work, minor fixes proposed.

new-client/src/plugins/Fir/FirImport.js
```diff
@@ -52,7 +54,6 @@ export default class FirImport {
   };
 
   translateImportedFeature(feature) {
-    // got this from the Draw plugin...
     var coordinates = feature.getGeometry().getCoordinates(),
```
Hallbergs, line 57: `const coordinates = ...`?

Hallbergs (changes requested):
I agree with @jacobwod comments.

new-client/src/buildConfig.json
jesade-vbg: Indeed, VS Code lies to me :)

new-client/src/plugins/Kir/KirLayerController.js
jesade-vbg: This is a matter of me being too old school, using old syntax. Now it's time to punch this "new" syntax into the spine.

new-client/src/plugins/Kir/KirSearchResultItemView.js
jesade-vbg: I dont agree with the best practice but I'll get it done :)

new-client/src/plugins/Kir/KirSearchView.js
jesade-vbg: I might keep this as is, and get back to it later after launch if that's ok.

new-client/src/plugins/Fir/FirImport.js
jesade-vbg: Indeed, it's a copyNpaste bonus. I'll fix!

jesade-vbg:
Thanx for reviewing! I'll fix everything but the "theme mui 5 comment" right now i't thats ok.
It's great to have more eyes to find these things, It's easy to get a bit "code blind".

Hallbergs (approved):

PR #982, opened 2022-01-14
Fix: AD search loop exits if one container is not found

maan002:
If you specify more than one AD container in Web.config and one is not found the loop exits.

This fix will continue to search all specified constainers even if one or more is not found.


Hallbergs (approved):
LGTM

PR #983, opened 2022-01-15
New FeatureInfo Filter Function lt

maan002:


Hallbergs (approved):
LGTM. Saw that the documentation was already updated, good job!

PR #1013, opened 2022-02-23
Dockerfile for UBI Nodejs

johwes:
I've created a Dockerfile that uses Red Hat's Universal Base Image (UBI) as the base image. I believe it will be of use to a wider audience that want to run on UBI.


Hallbergs (approved):
Nice addition! I'm thinking that we could move all the Dockerfiles to /Docker, but i can do that later.

PR #1017, opened 2022-02-28
Add USER root for builds to work on Openshift. Use minimal image

johwes:
Builds on Openshift needs a USER entry in the Dockerfile,
Also add the minimal image for the runtime part to reduce the final image size


Hallbergs (approved):

goteborgmiljo:
Excellent, @johwes!

Issue #1060, opened 2022-04-25
Sketch: Create circle via radius


WCarlsson:
**Is your feature request related to a problem? Please describe.**
In the old draw-plugin it was possible to create circles by inputing a radius. 

**Describe the solution you'd like**
Implement the feature in the new draw-plugin. Easiest could be in a similiar way to the old version. If the object to be drawn is a circle a new voluntary parameter is shown where the user can write the desired radius. If no value is written the

Hallbergs:
Just a heads up (so no one else starts working on this issue): got my trainee working on this one.

PR #1063, opened 2022-04-27
Plugin(Sketch): Radius click creation 

Tolfx:
Issue relation: #1060 

This is a fix to issue #1060, now you can click to create a circle. And it will use the radius the user has inputed as show on the image below.
image


jacobwod (approved):
The code looks fine to me! 👍

Hallbergs (approved):
LGTM

PR #1066, opened 2022-04-28
Feature/1065 analytics support

### LINKED ISSUE #1065: Add support for different analytics solutions
# Documentation 

## The API
In addition to the typical page view, the following custom events are being tracked:
- `mapLoaded`: when a specific map is loaded
- `pluginShown`: when a plugin's window becomes visible
- `layerShown`: when a layer changes visibility to `true`
- `spatialSearchPerformed`: when a spatial search is performed
- `textualSearchPerformed`: when a text search is performed

## Example setup in Plausible
In your Plausible installation, ensure to add the previously stated event names as _Goals_:
Skärmavbild 2022-05-03 kl  15 00 36

This will result in a nice breakdown in the UI. The example below illustrates which plugins are most popular in a certain setup:
Skärmavbild 2022-05-03 kl  15 01 17

# Original issue
Currently many organisations that wish to collect statistics about their Hajk usage do it by adding tracking snippets to the static HTML file when deploying. This works but adds a step to the deployment step and wouldn't work at all in a CI/CD pipeline.

I propose a different solution that would allow for better integration and offer even more possibilities. 

The idea is that we integrate one or more different (open source) tracking solutions and allow map admins to control whether they're active using Admin UI.

I will start by adding support for Plausible (https://plausible.io) which is open source, lightweight and allows for self hosting. It is easy to setup due to [ready Docker compose file](https://github.com/plausible/hosting/blob/master/docker-compose.yml). In addition, the [`plausible-tracker` package](https://www.npmjs.com/package/plausible-tracker) will allow us to import it programatically - no need to put `<script .... />` in our `index.html`!

jacobwod:
Closes #1065

### DIFF
--- a/new-backend/.env
+++ b/new-backend/.env
@@ -148,3 +148,12 @@ AD_PASSWORD=pass
 # *** OPENAPI SETTINGS ***
 OPENAPI_SPEC=/api/v1/spec
 OPENAPI_ENABLE_RESPONSE_VALIDATION=false
+
+# *** ANALYTICS SETTINGS ***
+# Type of service. Currently "plausible" is supported. Support for "matomo" is planned too.
+#ANALYTICS_TYPE="plausible" 
+# Options that will be passed to the service. Must start with "ANALYTICS_OPTION_". Please
+# refer to the specific service's documentation (https://github.com/plausible/plausible-tracker#usage)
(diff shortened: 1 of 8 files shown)


Hallbergs (approved):
Very nice

PR #1071, opened 2022-05-03
Option to disable autocomplete

### LINKED ISSUE #1070: Provide admin-option to disable autocomplete in search
In some Hajk-setups the autocomplete-functionality in Hajk seems to do more harm than good. For instance, some users seems to believe that the autocomplete-list is the actual search-results (making them a bit disappointed when they don't find what they are looking for).

Let's provide an admin-option that can be used to disable the autocomplete, and show the actual results right away instead.

With autocomplete active: (We show the autocomplete after around 300ms)
autocomplete_search

With autocomplete disabled: (We jump straight to the results after around 300ms)
regular_search

Hallbergs:
Closes #1070. See desc in issue.

### DIFF
--- a/new-admin/src/views/tools/search.jsx
+++ b/new-admin/src/views/tools/search.jsx
@@ -29,6 +29,7 @@ const defaultState = {
   scale: "",
   markerImg: "",
   delayBeforeAutoSearch: 500,
+  disableAutocomplete: false,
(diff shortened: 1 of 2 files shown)


jesade-vbg:
Great! It works really well when I tested.

jesade-vbg (approved):
Thumbs up!

PR #1073, opened 2022-05-04
Added `crossorigin="use-credentials"` on manifest-files. Closes #1072

### LINKED ISSUE #1072: Hajk behind authentication proxy =>  manifest.json 401:s
If Hajk is deployed behind an authentication proxy, requests for the `manifest.json` fails with a 401.

image

My guess is that this stems from the fact that no crossorigin-attribute is set on the link. Adding `crossorigin="use-credentials"` seems to fix the error.

Suggested fix: Add `crossorigin="use-credentials"` on the manifest-links (both in client and admin).

Hallbergs:
Some users has noticed that the `manifest.json` files fails to load when the solution is deployed behind an authentication proxy. Adding `crossorigin="use-credentials"` should solve the issue.

See #1072 for more info.

### DIFF
--- a/new-admin/public/index.html
+++ b/new-admin/public/index.html
@@ -9,7 +9,11 @@
       rel="stylesheet"
       type="text/css"
     />
-    <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
+    <link
+      rel="manifest"
+      href="%PUBLIC_URL%/manifest.json"
+      crossorigin="use-credentials"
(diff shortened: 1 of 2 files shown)


jacobwod:
Is this treated differently because of the MIME type? You haven't seen the same problems with the remaining `<link>` resources (logo192.png, index.css)?

jacobwod (approved):

PR #1074, opened 2022-05-04
Fix :hammer: `customGetMapUrl` used for legend. Closes #1062

### LINKED ISSUE #1062: LegendGraphic doesn't show on layers that use "GetMap-url" parameter
On layers that use different url:s, see https://github.com/hajkmap/Hajk/issues/345, for GetCapabilities and other requests the legendGraphic doesn´t show.

This seems to be because the GetLegendGraphic uses the wrong url for the request, for example:
Url: https://resource.sgu.se/service/wms/130/jordarter-25-100-tusen-fastmark
GetMap-url: https://maps3.sgu.se/geoserver/jord/ows
GetMap and GetFeatureInfo uses the GetMap url but getLegendGraphic uses the normal url.

1. Set up a layer with different url:s
2. Add layer to a map
3. Expand to see legendGraphic

**Expected behavior**
Legend Graphic should show as intended

Hallbergs:
Should take care of #1062

### DIFF
--- a/new-client/src/utils/ConfigMapper.js
+++ b/new-client/src/utils/ConfigMapper.js
@@ -53,7 +53,15 @@ export default class ConfigMapper {
         // to append the remaining values with &.
         const theGlue = args.url.includes("?") ? "&" : "?";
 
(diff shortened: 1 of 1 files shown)


jacobwod:
👏

jacobwod (approved):

Issue #1075, opened 2022-05-04
Sketch: Allow users to select features from other layers


Tolfx:
**Is your feature request related to a problem? Please describe.**
Right now is not possible to select features from layer and copy them, thus adding in sketch a draw method to copy them.

Tolfx:
Once copied, it will use the styles and be placed where the feature coordinates where as well, for this screenshot I moved it to show.
![image](https://user-images.githubusercontent.com/57797792/166651918-2744a0b0-dd0c-497b-9312-e7ad67e4acc5.png)

And if there are multiple layers it will have a pop of which one you would like to copy.
![image](https://user-images.githubusercontent.com/57797792/166652224-519cd4f8-5183-4c23-b094-0dd98200d388.png)

PR #1079, opened 2022-05-05
Plugin(sketch): Copy features/draw layer.

Tolfx:
Issue relation: #1075

Fix for issue #1075.

Will make it possible to click on a feature and copy and turn it into a sketch.

It will highlight which feature you want to copy and turn into a sketch.
image
image

Once applied it will be placed on top of the feature like this:
image

Also possible to copy the sketch like this:
image


Hallbergs (approved):
LGTM

PR #1134, opened 2022-07-20
Add options and MUI5 upgrades in Geosuite plugin

linusfj:
The following changes are included:
- To decide whether the options in the second step are selectable or not, two new admin options have been added for the Geosuite plugin.
- The styling has been rewritten, and the package sources have been changed, to work with MUI5.


Hallbergs (approved):
I'm getting a `controlled` change error, see below: 
image
I'm guessing this is due to me missing a proper plugin-configuration.

Since this is a plugin used only by a few, and the changes does not affect the rest of the application, i'm approving.

PR #1138, opened 2022-07-29
Upgrade UI for Coordinates plugin

### LINKED ISSUE #1114: Coordinates plugin: UI changes
The current UI could be improved with the following changes:

- button order
- button styling
- less margin
- text size

image

linusfj:
A few UI improvements have been made for the Coordinates plugin to make it more consistent to the other plugins.

Includes the following changes:
- Button order
- Buttons fill the viewport width on mobile devices
- Input fields fill the viewport width on mobile devices
- Less margin around elements
- Replace Tables with Grid components
- Use primary theme color on buttons


Hallbergs (changes requested):
First of all, i am glad that the coordinate plugin is getting reworked. The previous UI was not very nice.

There seems to be some issues with this implementation though... See below:

1. Alignment. Would look better to have the buttons centered? With the width set in `simpleMapConfig` we get this appearance: 
image 
If for some reason someone want a really wide coordinate plugin, we get this appearance: 
image
To make it look good you have to use a "perfect" width. Not desirable. 

2. If i click somewhere to see the current coordinate i get the following appearance: 
image
(Un-scrollable, quite hard to use tbh.)

new-client/src/plugins/Coordinates/CoordinatesView.js
```diff
@@ -81,53 +91,67 @@ class CoordinatesView extends React.PureComponent {
...
+                  this.props.model.goToUserLocation();
+                }}
+              >
+                Min position
+              </StyledButton>
+            </Tooltip>
+          </Grid>
+          <Grid item xs>
+            <Grid container direction="row-reverse">
+              <Grid item>
```
jacobwod, line 127: The selected rows are just an example - this comment is generic about the entire Grid you have in this View.

It looks to me as you try to have a [nested grid](https://mui.com/material-ui/react-grid/#nested-grid) but failed to created it correctly. So you end up with something like:
```
| item | item                 |
| another Item | another Item |
```
where you actually wanted:
```
| item         |  item        |
| another Item | another Item |
```

Take a look at MUI's documentation and how to create nested grids.

Hallbergs (approved):

PR #1140, opened 2022-08-16
Added possiblity for the user to set a custom print-scale.

Hallbergs:
For some users, the predefined scale-values doesn't fit their use case. This additions allows the users to select a custom scale.

image


jesade-vbg (approved):
Looks good! A great addition. I know some people in Varberg has asked for this feature as well.

jacobwod (approved):
Perfect! 👍

Issue #915, opened 2021-11-19
OpenLayers HIDPI often causes GeoServer cache to fail.


jesade-vbg:
Turning on HIDPI changes the way OpenLayer requests tiles etc.

You can specify diffrent dpi:s in geoserver, 90, 180 etc.
**But** OpenLayers requests the exact dpi as the device has, for example 72, 113, 155, 189 etc etc.
Confirmed in docs: HiDPI >
>  Use the ol/Map#pixelRatio value when requesting the image from the remote server.

If you're lucky, it requests 180 as it's set up in GeoServer.

For example, OpenLayers adds this to the Url:
```
WIDTH=538
HEIGHT=538
FORMAT_OPTIONS=dpi:189
```

Resulting in this header from geoserver:
`geowebcache-cache-result: MISS`
=== slow

Could we somehow force OpenLayers to request (as an example)?:
```
FORMAT_OPTIONS=dpi:90
WIDTH=256
HEIGHT=256
```
or double resolution for hidpi devices:
```
FORMAT_OPTIONS=dpi:180
WIDTH=512
HEIGHT=512
```
Everyone wants fast maps and sharp "crisp" maps. It would be nice to be able to deliver both.

jesade-vbg:
Spontaneous thought:
Combine custom imageLoadFunction (or similar) with window.devicePixelRatio

jesade-vbg:
**Quick test**: Turned off HiDPI and Forced 180dpi 512x512 using tileLoadFunction on the left. (Chrome device emulator).

![image](https://user-images.githubusercontent.com/73874338/142637412-58d3bbec-86c2-42a8-baeb-9ab2217c2241.png)

Hallbergs:
I think the idea of forcing specific DPI:s is good! Would 90 and 180 be enough (one for ordinary devices and one for highDPI-devices)? Or do we want more?

jesade-vbg:
Maybe it could be configured like (devicePixelRatio: dpi):
 `{
  "1.0": 90,
  "2.0": 180,
}`

Hallbergs:
Ahh very nice! I like it!

jesade-vbg:
POC
https://github.com/hajkmap/Hajk/tree/feature/915-custom-hdpi

Hallbergs:
The code looks great, nicely commented and easy to understand. Will give it a go next week!

jesade-vbg:
[...]

PR #1141, opened 2022-08-17
Feature/915 custom hdpi

jesade-vbg:
As discussed in https://github.com/hajkmap/Hajk/issues/915

Tested and seems to be working fine with both .Net and Node backend.


jesade-vbg:
Note; No rush with this PR!

jacobwod (approved):
I had a look at this and the fact that you have `config.useCustomDpiList` makes it quite safe to merge for us. (We won't be activating this as we have not run into the issues you solve in this PR - somehow the 90/180 conversion seems to work for the devices I've tested and I always get HIT from GWC.) 

Anyway, looks good to me and ready for merge, if you are confident about the actual functionality.

PR #1164, opened 2022-09-07
Fix/1115 admin wms projection

### LINKED ISSUE #1115: Problem adding some layers from GetCapabilites
**Describe the bug**
In _admin_, there is a problem to add WMS services from https://www.lantmateriet.se/sv/geodata/vara-produkter/produktlista/#category=visning.

**To Reproduce**

1.	Add new layer with following values:
a.	Server type: GeoServer.
b.	URL: Use any WMS service from https://www.lantmateriet.se/sv/geodata/vara-produkter/produktlista/#category=visning.
c.	Press button to load layer.
2.	Coordinate system dropdown menu is empty.
3.	Same problem occurs when changing server type.
4.	Save configuration and add layer to map.
5.	Try to activate the layer in _client_.

**Expected behavior**
In _admin_, projections in the coordinate system dropdown menu should be extracted from _GetCapabilities_.

linusfj:
Some servers return the projections in lower case, which causes an issue when trying to load the layer.

Upper case all projections to match the EPSG codes in the projection list. Closes #1115.

### DIFF
--- a/new-admin/src/views/layerforms/wmslayerform.jsx
+++ b/new-admin/src/views/layerforms/wmslayerform.jsx
@@ -1075,6 +1075,12 @@ class WMSLayerForm extends Component {
       projections = this.state.capabilities.Capability.Layer[RS];
     }
 
+    if (projections) {
+      projections = projections.map((projection) => {
+        return projection.toUpperCase();
+      });
+    }
+
     let projEles = projections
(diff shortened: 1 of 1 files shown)


jesade-vbg (approved):
Cant see any problems with this one

PR #1166, opened 2022-09-08
Feature/1160 sketch replace warning

OlofSvahnVbg:
Added a replace warning to SaveView in Sketch that prompts when user types a sketch name that already exists.

### DIFF
--- a/new-admin/src/views/edit.jsx
+++ b/new-admin/src/views/edit.jsx
@@ -229,16 +229,12 @@ class Edit extends Component {
   /**
    *
    */
-  getLayersWithFilter(filter) {
+  getLayersWithFilter() {
     return this.props.model.get("layers").filter((layer) => {
-      return (
-        new RegExp(this.state.filter.toLowerCase()).test(
-          layer.caption.toLowerCase()
-        ) ||
-        new RegExp(this.state.filter.toLowerCase()).test(
-          layer.internalLayerName?.toLowerCase()
-        )
-      );
+      const caption = layer.caption.toLowerCase();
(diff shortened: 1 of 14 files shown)


jesade-vbg:
@Hallbergs Hi, what is going on with the "All checks have failed"? Any idea?

Hallbergs:
There is an error with the Hajk version (3.11-dev). Netlify only accepts x.x.x-versions. I thought @jacobwod fixed this in develop. Hmm...

Hallbergs:
It should be fixed now. Will take a look at the pr when i get a chance

Hallbergs:
Hmm... I saw that some caption-typographys has been changed to spans. Any reason why? Not sure the new look is better? See before and after images:

Before:
image

After:
image


Also, i'm not sure the spacing under the name-input is necessary?
image

OlofSvahnVbg:
The caption change in the saved skethes list is a mistake, I think. Completely agree it looks better in the before image. The spacing was added so the text would appear more "smoothly" but can be removed easily (inline-block).

Hallbergs:
Alright! I see. I think it would look better if the spacing under the input is increased/decreased depending on if the warning text is shown or not! :) 

The addition seems to work nicely though! Nice job, will merge when styling is fixed.

OlofSvahnVbg:
Thanks. I'll make the changes asap.

OlofSvahnVbg:
Made commit but I found a bug that I'm currently working on.

Hallbergs (approved):
Good!

PR #1169, opened 2022-09-12
Fix automatic zoom on mobile devices

linusfj:
If you use the Safari browser on mobile devices, it automatically zooms in whenever you click on a text field.

To prevent the browser from doing that we can add `maximum-scale=1`, which controls how much zoom in is allowed on the page.

### DIFF
--- a/new-client/public/index.html
+++ b/new-client/public/index.html
@@ -2,7 +2,10 @@
 <html lang="sv">
   <head>
     <meta charset="utf-8" />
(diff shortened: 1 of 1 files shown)


Hallbergs (approved):
Should work! I haven't tested it on an actual mobile device though. 

I've seen the issue with mobile zoom on text-inputs with small text-size, we solved it by increasing the text size to the recommended size. This is still a good addition i think (the app should never be used zoomed).

PR #1170, opened 2022-09-13
Added support to mimic proxyBaseUrl in GeoServer

maan002:


jesade-vbg (approved):
Works as intended, good work!

PR #1173, opened 2022-09-14
Feature/1093 new dummy plugin

Hallbergs:
Updated Dummy-plugin based on functional components.

Includes:
- [X] Lots of information and tips
- [X] Updating local-storage (also making sure we're allowed to)
- [X] Using core models
- [X] and more...


jesade-vbg (approved):
Looks really awesome! Should help alot, both to get coders to produce and to get them to make it right (follow best practice). I also like the Object.freeze (dont mess with this stuff) :)

PR #1183, opened 2022-09-22
Feature/1174 print marginal

### LINKED ISSUE #1174: Print: add an option to append a margin
User has asked to increase the marginals in the pdf-print function. Reason being; the icons and text now overlaps the map image, which renders them harder to discern. Marginals should be increased to fit the icons and text, so that they do not interfere with the map area:

This should be an option that can be turned on/off.

PDF with upper and lower margins

OlofSvahnVbg:
Closes #1174

### DIFF
--- a/new-client/src/plugins/Print/PrintModel.js
+++ b/new-client/src/plugins/Print/PrintModel.js
@@ -1064,7 +1064,7 @@ export default class PrintModel {
       // Add map title if user supplied one
       if (options.mapTitle.trim().length > 0) {
         let yPos = options.useTextIconsInMargin
-          ? this.margin * 2
+          ? 8 + this.margin
           : 12 + this.margin;
         pdf.setFontSize(24);
         pdf.setTextColor(options.mapTextColor);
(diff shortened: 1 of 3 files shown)


new-client/src/plugins/Print/PrintModel.js
```diff
@@ -953,13 +970,21 @@ export default class PrintModel {
...
       // Add potential margin around the image
       if (this.margin > 0) {
-        // The lineWidth increases the line width equally to "both sides",
-        // therefore, we must have a line width two times the margin we want.
-        pdf.setLineWidth(this.margin * 2);
         // We always want a white margin
         pdf.setDrawColor("white");
-        // Draw the border (margin) around the entire image
-        pdf.rect(0, 0, dim[0], dim[1], "S");
+        // The lineWidth increases the line width equally to "both sides",
```
Hallbergs, line 975: Move this comment so that it makes sense and try to describe whats going on with A5 case.

OlofSvahnVbg: Sir, yes sir!

OlofSvahnVbg:
Fixed the comments for a5 and added a small adjustment to preview window, also for a5 format.

Hallbergs (approved):
Very nice!

PR #1189, opened 2022-09-27
Fix/1181 cyrillic print support

Hallbergs:
Updated the font used when printing to support cyrillic characters. Decided to use the Roboto font which complies with the rest of the application.

Further development would be to allow the user to select font, but thats a separate issue.


jacobwod (approved):
Fantastic. I also like the fact that even Latin alphabet users benefit from this as we now get the same font in our PFDs as in the application itself. Nice!

jesade-vbg:
Good work. Great that Roboto is used instead.

jesade-vbg (approved):

PR #1190, opened 2022-09-27
Potential fix for #1186

### LINKED ISSUE #1186: Sketch: make it possible to disable the "halo" on text nodes
Currently user must select both foreground and background color when adding text via Sketch.

We should make it possible for users to select "None" for the background, so no halo is generated. There are situations where this is desirable. 

A simple fix (selecting the same color for both options) renders a hard-to-read label, so it is unfortunately not an option.

Hallbergs:
Makes it possible to create text features in sketch where the outline is removed if it has the same color as the foreground.

Closes #1186


jacobwod (approved):
Great solution, as usual! 👍

jesade-vbg:
Looks good!

jesade-vbg (approved):

PR #1192, opened 2022-09-29
Feature/1191 placement warning

### LINKED ISSUE #1191: Print: Placement of the warning label
There should be a warning in "Print" when the user selects overlapping northarrow/scalebar/logo. As it is now, no warning is given when the user, for example, selects top right placement for all three icons.

The user should get a warning underneath the selections like this:

Skärmbild 2022-09-28 085356

Of course, the user should still be able to overlap the icons if they wish.

OlofSvahnVbg:
closes #1191


Hallbergs (approved):
Great!

PR #1194, opened 2022-09-29
Control buttons render in Drawer on small screens, closes #1193.

### LINKED ISSUE #1193: Proposal: all plugins (regardless of `target`) render in Drawer on small screens
Currently Hajk can be configured in such a way that the number of control buttons (i.e. plugins where `target: "control"` is set) is so big that we run out of vertical space. See also #1171 where this issue first materialised, especially [here](https://github.com/hajkmap/Hajk/issues/1171#issuecomment-1260424522).

I propose a change of current functionality that mimics what we currently do with Widget buttons: on small viewports we "move" those buttons from their normal position into the Drawer (sidebar panel). 

I think this should be done for control buttons too.

jacobwod:


Hallbergs (approved):
Looks good!

Hallbergs:
@jesade-vbg What do you think about this one?

jesade-vbg (approved):
Good work!

PR #1199, opened 2022-10-03
Added date prop to print

### LINKED ISSUE #1196: Print: Add current date to printout
A user has requested we add a date to Print. Preferably the date should be printed where the "disclaimer" and "copyright" is on the printed image (see red marking in image below).

Skärmbild 2022-09-29 135858

OlofSvahnVbg:
closes #1196

### DIFF
--- a/backend/mapservice/App_Data/map_1.json
+++ b/backend/mapservice/App_Data/map_1.json
@@ -457,6 +457,7 @@
     {
       "type": "print",
       "options": {
+        "date": "Utskriftsdatum: {date}",
         "target": "left",
         "position": "left",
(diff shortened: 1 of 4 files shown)


OlofSvahnVbg:
Havent thought about adding to admin, but would probably be easy.

OlofSvahnVbg:
I checked with colleagues regarding date in admin, but no one knew where to add this. Maybe we can check on that later on?

Hallbergs:
> I checked with colleagues regarding date in admin, but no one knew where to add this. Maybe we can check on that later on?

Yes, sure. I would say we should add it to the print-tool-section in admin? Otherwise there will be no way to enable the date in print without messing with the config files manually?

OlofSvahnVbg:
> 

Agreed.

jesade-vbg:
I totally forgot to mention the needed Admin changes to Olof. We will create a new Issue with the needed changes and assign it to us.

PR #1201, opened 2022-10-05
1200 admin marginals and date

### LINKED ISSUE #1200: Add increased marginals and date printout to admin
We should add "larger marginals" ( #1174 ) and "date printout" ( #1196 ) to admin. I guess that's it. See example for "date" below:

image

And example for larger margins below: 

image

### LINKED ISSUE #1203: Shrink north arrow
The north arrow should be made smaller. When the user selects "Rubriktext m.m. i marginalerna" the map image is cropped to fit the height of the north arrow, and subsequently the map image gets a bit "thin". To avoid this the north arrow could be made smaller, allowing the map image to be taller.

image

image

OlofSvahnVbg:
Closes #1200
Closes #1203

### DIFF
--- a/new-admin/src/views/tools/print.jsx
+++ b/new-admin/src/views/tools/print.jsx
@@ -24,6 +24,7 @@ var defaultState = {
   instruction: "",
   copyright: "",
   disclaimer: "",
+  date: "",
   scales: "200, 400, 1000, 2000, 5000, 10000, 25000, 50000, 100000, 200000",
   dpis: "72, 150, 300",
   paperFormats: "A2, A3, A4",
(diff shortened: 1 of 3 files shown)


Hallbergs (dismissed):

jacobwod:
I think this is great. The only problem I run into is with our logo (which seems bigger than the one you use?). Can we handle it somehow, but shrinking if some size is exceeded, similarly to how you've done it with the north arrow in #1203?

jacobwod (dismissed):
Looks fine but please see separate comment considering the logo size.

OlofSvahnVbg:
@jacobwod We decided on adding a northArrowMaxWidth for various reasons. Some users might be inclined to have different max values for logo and north arrow, depending on preference.

Hallbergs (approved):
lgtm. Merge when you feel ready!

PR #1205, opened 2022-10-05
Fix/1204 infoclick title

### LINKED ISSUE #1204: Wrong title for LayerGroup in infoclick
**Describe the bug**
When _infoclicking_ on sublayers of the type `LayerGroup`, the incorrect caption is displayed in the title.

**To Reproduce**

1. Add layer as a `LayerGroup`.
2. _Infoclick_ on a sublayer in the group.
3. The title does not use the correct caption in the information window.

linusfj:
Closes #1204.

### DIFF
--- a/new-client/src/components/FeatureInfo/FeatureInfoContainer.js
+++ b/new-client/src/components/FeatureInfo/FeatureInfoContainer.js
@@ -220,10 +220,11 @@ class FeatureInfoContainer extends React.PureComponent {
     const layerInfo = feature.layer.get("layerInfo");
 
     // Get current id of feature and find out if it occurs in the layersInfo array.
-    let featureId = feature.getId();
+    // Remove the unique identifier after the last dot, since there can be dots that is part of the ID.
+    let featureId = feature.getId().split(".").slice(0, -1).join(".");
     let layersInfo = layerInfo.layersInfo;
-    let layerId = Object.keys(layersInfo).find((key) =>
-      featureId.includes(layersInfo[key].id)
+    let layerId = Object.keys(layersInfo).find(
+      (key) => featureId === layersInfo[key].id
     );
 
     let markdown = layerInfo?.information,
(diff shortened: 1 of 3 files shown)


jacobwod (changes requested):
Could not reproduce the original issue, please see my comments in #1204.

jacobwod:
We've looked into this. The proposed solution (in this PR) will unfortunately not work as it can lead to false positive matches (due to the nature of `.includes()`).

Imaging this very real example of a group layer consisting of sublayers:

```javascript
"layers": [
                    "fastighetsgranser",
                    "fastighetsytor",
                    "fastighet",

                ],
```

The proposed solution would bring the same `caption` for all the first and last sublayer because the loop that looks for matching sublayer would evaluate to true two times. (`"fastighet".includes("fastighet") === true` as well as `"fastighetsgranser".includes("fastighet") === true`). 

So we need more robust logic in this lookup.

jacobwod:
To sum up the previous post: this solution seems to work but is unfortunately limited to only fix the issue in the "old" infoclick variant. I'm not sure I like to see Hajk diverge this much - it's one thing to add new features that are not backported, but I think that fixing bugs should be done for the entire solution. 

Especially since it's easy for an admin to deactivate this fix accidentally, as it's only one checkbox away…
image


What's your take on this @Hallbergs, @jesade-vbg?

Hallbergs:
> To sum up the previous post: this solution seems to work but is unfortunately limited to only fix the issue in the "old" infoclick variant. I'm not sure I like to see Hajk diverge this much - it's one thing to add new features that are not backported, but I think that fixing bugs should be done for the entire solution.
> 
> Especially since it's easy for an admin to deactivate this fix accidentally, as it's only one checkbox away… image
> 
> What's your take on this @Hallbergs, @jesade-vbg?

I agree that the best solution would be to solve the issue in both the new and old feature-info. @linusfj do you have time to fix it in the new one as well? If not, please make an issue stating that the new one still got the group-bug. (And we can make sure that it is fixed before the next release).

linusfj:
I’ll try to fix it in the new _infoclick_ as well and update this pull request.

jacobwod:
I've merged your latest changes and tested, but I'm not sure that you verified how it works with the new infoclick. I'll see what I can do to fix this:
Skärmavbild 2022-11-03 kl  09 42 30

jacobwod:
You forgot to pull in the lastest `develop`. It seems to look good now.

jacobwod (approved):

Issue #987, opened 2022-01-26
Can't save text field in Edit


jacobwod:
This needs more investigation, but to reproduce, try:
1. Setup a new WFS-T layer. Mine looks like this:
![Skärmavbild 2022-01-26 kl  10 37 04](https://user-images.githubusercontent.com/110222/151138984-61ff8f85-227b-4910-8f5f-082945dc5e6b.png)

2. Add it to a map. 
3. Try adding a point. First without supplying any data in the text field. 
4. **This saves the geometry correctly in the database**. 
![Skärmavbild 2022-01-26 kl  10 38 10](https://user-images.githubusercontent.com/110222/151139149-b32ab5f4-90e3-488d-b9aa-e76a9d91cf3b.png)
5. No try adding another geometry **with a text value this time**. It will fail, something like this:
![Skärmavbild 2022-01-26 kl  10 35 18](https://user-images.githubusercontent.com/110222/151138755-75297c39-fb99-47cd-bace-9307c8789c41.png)
6. You can also get the same error if you try to **edit the previously added geometry**. In addition, you will get `Object [object]` as the text field's value:
![Skärmavbild 2022-01-26 kl  10 39 30](https://user-images.githubusercontent.com/110222/151139362-1b6955e4-64d4-4e3f-ac08-859bb9a84b48.png)

hitomin21:
Hi, can you send the config you used for the layer and the plugin? I've still not been able to reproduce this so difficult for me to debug/fix. Am I correct that you use QGIS Server?

jacobwod:
I can confirm that the test instance works well and can easily be used as a remote backend even when working locally. Just change you `appConfig.json` to include this:
```
  "mapserviceBase": " https://hajk.orebro.se/api/v1/",
  "defaultMap": "map_1",
```

Well done @WCarlsson!

jacobwod:
I can confirm that the errors remain in the latest `develop`.
![Skärmavbild 2022-02-17 kl  07 32 34](https://user-images.githubusercontent.com/110222/154418866-e5e0f640-af76-4f9d-90f4-842aaedce2a5.png)

hitomin21:
Thanks, I will take a look

Anyou22:
Hi,
We are trying to solve this bug #987 but we are wondering If It has something to do with resetView(new-client, plugins, edit.js)?
[...]

PR #1228, opened 2022-11-01
Feature/orebro tsm

Sunix71:
Issue: https://github.com/hajkmap/Hajk/issues/987

QGIS server returns object that is not empty. HAJK prints out object as string which returns Object Object in fields (text och siffra).
Filename: AttributeEditor.js in new-client/src/plugins/Edit/components

### DIFF
--- a/new-client/src/plugins/Edit/components/AttributeEditor.js
+++ b/new-client/src/plugins/Edit/components/AttributeEditor.js
@@ -76,8 +76,15 @@ class AttributeEditor extends React.Component {
         } else {
           //If the feature has field: "" it will be changed to the default value.
           //Not sure if we want this behavior?
-          valueMap[field.name] =
-            featureProps[field.name] || field.defaultValue || "";
+          //Object that returns as a string results in [object] [Object]
+          if (
+            JSON.stringify(featureProps[field.name]) === '{"xsi:nil":"true"}'
+          ) {
+            valueMap[field.name] = "";
+          } else {
+            valueMap[field.name] =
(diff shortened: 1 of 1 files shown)


new-client/src/plugins/Edit/components/AttributeEditor.js
```diff
@@ -76,8 +76,15 @@ class AttributeEditor extends React.Component {
         } else {
           //If the feature has field: "" it will be changed to the default value.
           //Not sure if we want this behavior?
-          valueMap[field.name] =
-            featureProps[field.name] || field.defaultValue || "";
+          //Object that returns as a string results in [object] [Object]
+          if (
+            JSON.stringify(featureProps[field.name]) === '{"xsi:nil":"true"}'
```
jacobwod, line 81: To be honest, I don't see why you compare a stringified object, instead of just
```featureProps[field.name]?.["xsi:nil"] === true```

jacobwod (changes requested):
I'm not sure this simple fix is sufficient to close #987 but perhaps it's a step in the right direction. 

In addition, please see comment in code.

new-client/src/plugins/Edit/components/AttributeEditor.js
Sunix71: I'll write like this then:

featureProps[field.name]?.["xsi:nil"] === "true"
            ? (valueMap[field.name] = "")
            : (valueMap[field.name] =
                featureProps[field.name] || field.defaultValue || "");

Then it seems to work

jacobwod (approved):

PR #1230, opened 2022-11-04
Add close button to Snackbar in Coordinates Plugin

linusfj:
When minimizing the Coordinates Plugin on mobile devices, the window is hidden behind the Snackbar and it is difficult to maximize it.

Adding a close button to the Snackbar solves this problem by giving the user access to the maximize button again and without changing the placement of the Snackbar.

Before:
image

After:
image


jacobwod:
Can you please create a corresponding issue before you submit PRs? This way we could keep better track of ongoing development within the community.

PR #1241, opened 2022-11-14
Feature/786 grouplayer infoclick

### LINKED ISSUE #786: Handle InfoClick for group layers in Geoserver
_Hajk can show group layers from geoserver today but it's not possible to use InfoClick with correct behaviour._

**Expectations**
Each layer in a group should be able to have a specific html/markdown.
If no html/markdown is specified, it should fallback and use the parents html/markdown.

**Why make this possible?**
If we were able to group layers in geoserver and still get all needed feature information,
and format the information correctly it could drastically improve performance by minimizing requests.
In some scenarios this would be great.

_Hajk has the ability to group layers but the request count are still unaffected._

Do you have any other suggestions/ideas? let me know.

jesade-vbg:
Closes #786


jacobwod:
Interesting, will take a look right away!

jacobwod:
Ah sorry, I see that it's just a draft so far! I guess I was too 🏃💨

jesade-vbg:
@jacobwod well it was ready, I just forgot to toggle it :)

PR #1242, opened 2022-11-14
Feature/1227 verify layers

### LINKED ISSUE #1227: Backend: Add an endpoint that verifies Hajk layers against their respective service
As a Hajk setup grows, it can easily consist of thousands of Hajk layers (layers defined in `layers.json`). Over time, it because impossible to keep track of each layer individually. 

The respective service (WMS, WFS, etc) can move the layer to another workspace, rename it or just become unavailable. **All these situations will lead to Hajk Client having problem when loading maps with layers that do not exist where they are expected.**

To ensure a high quality of services provided by Hajk, Backend could be extended with an endpoint that checks each layer's existence. The exact implementation will vary, but the main idea is to:
1. loop each of the defined layers
2. group them together depending on source 
3. ask each source for a list of available layers (usually `GetCapabilities`)
4. compare the response from source with what we have in `layers.json`
5. mark diff layers as missing and return with the response

jacobwod:
Please take a quick look. The code is pretty isolated and I've ensured to do some checks so it won't crash arbitrarily, but feel free to leave comments!

Closes #1227


jesade-vbg:
**Note that you'll need latest LTS of node 18+.**  
_Error: Function not supported. Please update your NodeJS runtime to at least v18.0.0. Current Node version: v16.16.0._

jesade-vbg (approved):
Looks good Jacob! Good addition

jacobwod:
> **Note that you'll need latest LTS of node 18+.** _Error: Function not supported. Please update your NodeJS runtime to at least v18.0.0. Current Node version: v16.16.0._
> 
> Just noticed in the code that this is well known :) But I think this is worth noting so people just dont deploy live and notices this later.

Yes, I thought that the best way is to allow the process to run on older Node runtimes but display this message. It's not a mission critical in my opinion, so we want it to run anyway, but it's good to hint system admin that an update is welcome. :)

PR #1246, opened 2022-11-17
Added group layers support to Anchor

### LINKED ISSUE #1226: Anchor: generated link enables all sublayers even if only a subset was selected initially
**Describe the bug**
The Anchor plugin (which really should be renamed to _Share_ by the way) has trouble working with Hajk's group layers.

Even if a group layer is only partly selected, the Anchor plugin's will generate a URL that will lead to all sublayers being active:

https://user-images.githubusercontent.com/110222/197768172-2bef8735-6a8d-41f8-99d6-e0e64586f4a2.mov

jacobwod:
Closes #1226

### DIFF
--- a/new-client/src/models/AppModel.js
+++ b/new-client/src/models/AppModel.js
@@ -45,6 +45,7 @@ class AppModel {
     this.plugins = {};
     this.activeTool = undefined;
     this.layersFromParams = [];
+    this.groupLayersFromParams = [];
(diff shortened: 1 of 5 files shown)


jacobwod:
Good to go @jesade-vbg, @Hallbergs. I think I've covered all possible cases but feel free to test and see if you can get this to crash. 😉

Hallbergs (approved):
Woho! Works great! :D

PR #1249, opened 2022-11-29
Feature/1176 sketch control zindex

### LINKED ISSUE #1176: Sketch: make it possible to control the zIndex of features
Currently the features zIndex seems to be determinate by creation time: most recent features are inserted at top of the stack.

It would be very useful if the user could control this somehow. One way is to increment the zIndex during an edit/move operation, so that the edited feature ends up on top. Another option could be a button inside the Edit window (something like "Move to top" or similar). 

image

jesade-vbg:
closes #1176

### DIFF
--- a/new-client/src/models/DrawModel.js
+++ b/new-client/src/models/DrawModel.js
@@ -326,6 +326,7 @@ class DrawModel {
     }
   };
 
+  // Find the highest zindex used and move on top of it.
   moveFeatureZIndexToTop = (feature) => {
     const indexes = this.#getAllZIndexes();
     const topIndex = Math.max(...indexes);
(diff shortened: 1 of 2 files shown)


jacobwod (dismissed):
Checked out, tested. Love the functionallity!

Hallbergs:
I'm gonna have a quick look, wait a minute before merging.

Hallbergs (changes requested):
If you save a sketch (with the LS save, seems to work with .kml), the order is changed (back to the original drawing order i think). Could you have a look at that?

jesade-vbg:
> If you save a sketch (with the LS save, seems to work with .kml), the order is changed (back to the original drawing order i think). Could you have a look at that?
> 
> Besides that it seems to work great!!

Oh no, it seemed to work fine when I tested, but now some scenerios seem "sketchy"

jesade-vbg:
@Hallbergs. When you have time, please test again.

Hallbergs (approved):
Works great! Well done :D

PR #1251, opened 2022-12-01
Added functionality for adding frame around image in admin

OlofSvahnVbg:
You can now select in admin to add a dark gray frame around the map image in the pdf. The option to add frame is only available there, and not for the user.

### DIFF
--- a/new-client/src/plugins/Print/PrintModel.js
+++ b/new-client/src/plugins/Print/PrintModel.js
@@ -27,6 +27,7 @@ export default class PrintModel {
     this.logoUrl = settings.options.logo || "";
     this.northArrowUrl = settings.options.northArrow || "";
     this.logoMaxWidth = settings.options.logoMaxWidth;
+    this.includeImageBorder = settings.options.includeImageBorder;
     this.scales = settings.options.scales;
     this.copyright = settings.options.copyright || "";
     this.date = settings.options.date || "";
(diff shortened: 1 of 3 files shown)


new-client/src/plugins/Print/PrintModel.js
```diff
@@ -1012,6 +1013,19 @@ export default class PrintModel {
           pdf.setLineWidth(this.margin * 2);
           // Draw the border (margin) around the entire image
           pdf.rect(0, 0, dim[0], dim[1], "S");
+          // If selected as feature in Admin, we draw a frame around the map image
+          if (this.includeImageBorder) {
+            // Frame color is set to dark gray
+            pdf.setDrawColor("#5A5A5A");
```
jacobwod, line 1019: Any reason why the color needs to be hard-coded? Perhaps we could use the same color as text color (which is user-selectable), or allow admins to set the frame color in Admin UI?

OlofSvahnVbg: I think we could use same color as text. Not sure if frame color needs to be selectable. Leaning towards not :)

jacobwod:
Nice!

Just one more thing: I don't want to be picky, but how does the frame affect the print area? I.e., when the frame is visible, do we reflect that in print preview (the "orange rectangle" that user can move around and select what will be printed). 

I know that this is just a minor addition but when you zoom in close enough, the width of the frame can affect the viewport substantially. We've had users complaining about "_Print_ does not print exactly what I selected!" before… 😄

OlofSvahnVbg:
Good point. I'll check on it.

OlofSvahnVbg:
image

image

OlofSvahnVbg:
The frame doesnt seem to obscure any map area, not that I can see at least. What do you think?

jacobwod:
It should, I think… :) Unless you've enabled `useMargin` in options, of course. That way the frame will end up inside the margin and not affect the map. But if `useMargin` is `false`, the frame _should_ obstruct some part of the map?

jacobwod:
Looks good @OlofSvahnVbg, it seems like you need to merge in `develop` as there's one conflict in PrintModel currently. 👍

jacobwod:
Just wondering if you and @jesade-vbg already had a look at this together, or if we should await @jesade-vbg's review?

jacobwod (approved):

PR #1255, opened 2022-12-07
Feature/1219 configurable drawer button

### LINKED ISSUE #1219: Making the documenthandler drawerButtonIcon configurable
We want to change the icon of the button used to open Dokumenthanterare 2.0.

We suggest making the documenthandler drawerButtonIcon configurable in admin. (in the same way that drawerTitle and drawerButtonTitle are already configurable)
The default icon is suggested to be MenuBook.

linusfj:
Closes #1219.


Hallbergs (changes requested):
It seems like the icon is placed a bit off, see image below:

image

The 8px margin you added is probably the culprit.

Hallbergs:
It is still off, check image below: 
image

Checking if `drawerButtonTitle` is nullish won't work since it is always set. We decide if we render it or not based on viewport size.

Also, just curious. Why an `<i>`-tag?

linusfj:
The icon has been updated to use a custom component instead, which has been styled to not add margin on some viewports.

Hallbergs (approved):

PR #1256, opened 2022-12-15
Feature/1099 infoclick parse xml

### LINKED ISSUE #1099: GetFeatureInfo - text/xml does not work in HAJK
We have connected a layer with "infoFormat": "text/xml" which should be a working format in HAJK(?) 
Despite this, we don't get an info click box when we info click on the layer.

https://ext-geodata-lokala.lansstyrelsen.se/arcgis/services/LSTO/lsto_wms_strandskydd/MapServer/WMSServer?SERVICE=WMS&VERSION=1.1.1&REQUEST=GetFeatureInfo&FORMAT=image%2Fpng&TRANSPARENT=true&QUERY_LAYERS=0&LAYERS=0&INFO_FORMAT=text%2Fxml&SRS=EPSG%3A3007&TILED=false&STYLES=&FEATURE_COUNT=100&X=50&Y=50&WIDTH=101&HEIGHT=101&BBOX=145698.01606499997%2C6393043.9795699995%2C148526.02515499998%2C6395871.98866

linusfj:
Closes #1099.

### DIFF
--- a/new-client/src/models/Click.js
+++ b/new-client/src/models/Click.js
@@ -178,8 +178,8 @@ function getFeaturesFromJson(response, jsonData) {
     parsed.forEach((f) => {
       f.layer = response.layer;
     });
-    const sortedFeatures = sortFeatures(response.layer, parsed);
-    return sortedFeatures;
+    sortFeatures(response.layer, parsed);
+    return parsed;
   } else {
     return [];
   }
@@ -192,8 +192,8 @@ function getFeaturesFromGml(response, text) {
     parsed.forEach((f) => {
       f.layer = response.layer;
     });
-    const sortedFeatures = sortFeatures(response.layer, parsed);
-    return sortedFeatures;
+    sortFeatures(response.layer, parsed);
+    return parsed;
   } else {
     return [];
   }
@@ -227,8 +227,8 @@ function getFeaturesFromXmlOrGml(response, text) {
     features.push(feature);
   }
 
-  const sortedFeatures = sortFeatures(response.layer, features);
-  return sortedFeatures;
+  sortFeatures(response.layer, features);
+  return features;
 }
 
 /**
--- a/new-client/src/models/Click.js
+++ b/new-client/src/models/Click.js
@@ -1,8 +1,9 @@
+import Feature from "ol/Feature";
 import GeoJSON from "ol/format/GeoJSON";
+import WMSGetFeatureInfo from "ol/format/WMSGetFeatureInfo";
 import TileLayer from "ol/layer/Tile";
 import ImageLayer from "ol/layer/Image";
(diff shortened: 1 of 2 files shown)


jacobwod:
Ready for review and merge @linusfj? In that case, please request a reviewer in the side panel.

jacobwod (commented):
A couple of questions follow.

Why do we need a custom XML parser here? Wouldn't [the way shown in OL example](https://openlayers.org/en/latest/examples/getfeatureinfo-layers.html) be sufficient enough? Is the response malformed in some way?

new-client/src/models/Click.js
```diff
@@ -275,7 +304,26 @@ export function handleClick(evt, map, callback) {
...
+                  })
+                  .catch((err) => {
+                    console.error(
+                      "GetFeatureInfo couldn't retrieve correct data for the clicked object.",
+                      err
+                    );
+                  })
+              );
+              break;
+            }
```
jacobwod, line 326: Also, this part looks very much like the next that follows, the only difference being which parser is used depending on XML or GML, i.e. lines 314 and 332. The rest is just the same so consider refactoring.

jacobwod:
Please confirm when you've tested the latest commits @linusfj.

linusfj:
It seems to not be working, when clicking on a feature an error message appears: 
`GetFeatureInfo could not retrieve the correct information for the selected object.`

jacobwod:
OK, take a look at what it is and try to fix it. I could not reproduce the error you mention on my side.

jacobwod:
OK, good find. Since we're now not using the return value form `sortFeatures`, could you please remove the `return` statement in that function? 

If you want to make it even nicer, you can rename `sortFeatures` to something like `sortAndMutateFeaturesArray` (but that's purely optional of course). 👍

jacobwod (approved):
Looks good!

jacobwod:
Btw, this is not part of this issue but I'll post it here: as I reviewed the code, it seems to look like the sorting algorithm only takes place in `Click.js`, while the new implementation in `MapClickModel.js` doesn't have the same sorting. Have you seen any difference in the results you get when toggling the `useNewInfoclick` flag in settings? If there is a difference, this will become an issue at some point in time, when those using the "old" infoclick choose to upgrade to the "new" method. If sorting is not implemented by then, the upgrade will have the side-effect of results being sorted differently. 

As I said, not part of this PR (it's ready to merge in my opinion). But I wonder what you think of it?

linusfj:
I've only tested the new infoclick version using a few layers with different data formats that we initially had problems with, and it was returning the correct results in those cases.

The new implementation in `MapClickModel.js` seems to be handling it well, but we should do more testing to ensure it works in all cases.

linusfj:
Merging pull request.

Issue #1258, opened 2022-12-22
Extend InfoClick support


Hallbergs:
While i was updating the config for our demo environments i found that some WMS providers returns strange responses from `getFeatureInfo`. See image below for an example:

![image](https://user-images.githubusercontent.com/43059093/209082702-e33821a6-b658-481d-9a9c-396581ce9f00.png)

These responses are not rendered correctly in the FeatureInfo-window since we expect the `id` to be constructed as
 `<layerName>.<some-numeric>` when we try to grab the necessary information for the FeatureInfo-window from config. In the case above, another random numeric is added to the id...

I am gonna try to sort this out, and hopefully this fix can be added to the 3.11 release. I'm a bit afraid it can't be added though, since it will probably be quite big changes since the logic around the FeatureInfo is...
![image](https://user-images.githubusercontent.com/43059093/209083703-c78a7bed-329e-4869-9e6c-eb8a334ffab1.png)

jacobwod:
Great initiative with fixing the config and fantastic upcoming solution. 

Let's see what it feels like after the Hollidays: whether we go for a quick merge to upcoming release or take some time to untangle this spaghetti.

PR #1262, opened 2023-01-09
Fix/1258 extend feature info support

Hallbergs:
Should take care of #1258 

Please test both the new and old infoClick with your local configurations. I've tested it with several layers without problems, but i've probably missed something.

Decided to target develop, it's a bit sketchy to put it in for 3.11 at this stage.

Henrik


jacobwod:
Looks good, but I'll check it out in detail and do the full review before the day is over!

jacobwod (approved):
No problems over here, works as intended. Good idea to refactor the helper to utils/.

jesade-vbg (approved):
Looks good! Tested a hand full of infoclicks and it worked fine. Hard to test them all. Good work!

PR #1270, opened 2023-01-20
Feature/1267 refactor bookmarks plugin

### LINKED ISSUE #1267: Refactor Bookmarks plugin to functional components
Update the Bookmarks plugin to use functional component implementations instead of class components, in order to align with the latest React best practices.

linusfj:
Closes #1267.

### DIFF
--- a/new-client/public/simpleMapConfig.json
+++ b/new-client/public/simpleMapConfig.json
@@ -439,7 +439,7 @@
       "index": 15
     },
     {
-      "type": "bookmark",
+      "type": "bookmarks",
       "options": {
         "target": "toolbar",
         "instruction": "",
--- a/new-client/src/plugins/Bookmarks/Bookmarks.js
+++ b/new-client/src/plugins/Bookmarks/Bookmarks.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState } from "react";
 import PropTypes from "prop-types";
 import BaseWindowPlugin from "../BaseWindowPlugin";
 import BookmarksModel from "./BookmarksModel";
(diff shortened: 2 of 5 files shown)


jesade-vbg:
Nice! It looks promising. I will check more and test next week. Had a quick check regarding storage format and it's indeed the parameters that are stored, including map name. Bookmarks are per domain not per map.

image

Hallbergs (changes requested):
I've just given it a quick look. Will give it a deeper dive next week. However, i have some questions right away:

 - Why are you rewriting and incorporating react functionality into `BookmarksModel.js`? This was just a regular js class, not a react class component (should/does not have to not be rewritten). Also i really don't think it a good idea with a react component(?) returning an object...? Or is this some kind of custom hook with bad naming? (Custom hooks should have the `use` keyword in front). I might just be too tired and missing something obvious. My main issue with this is that all our plugins are written with components along with one (or more) "model" (a regular js class). I don't see the point diverging from this "standard".
 - You "initialize" the model by writing `const bookmarksModel = BookmarksModel({ app, map });`. This code will run on every re-render of `Bookmarks.js` which is really unnecessary. You can look at the dummy-example where i thoroughly explain how this should be done.
 - `updateCustomProp` is unused and should be removed.

jesade-vbg:
I agree with Hallbergs comment. BookmarkModel.js should be kept as ordinary class, no reason to bloat it with react. I will look again and test when this is fixed.

jacobwod:
I agree with other reviewers. I think there's a clear example in the Dummy plugin. Otherwise check in e.g. Sketch.

jacobwod (changes requested):
I agree with other reviewers. I think there's a clear example in the Dummy plugin. Otherwise check in e.g. Sketch.

Hallbergs (changes requested):
I'm not sure if you're done? I checked the code again, and found some issues.

First of all, make sure we're compiling without warnings. See image below:
image

The warnings stems from the fact that you've added stuff (from `Dummy`?) that are unnecessary, For example the `pluginShown` state was added in `Dummy` to show how to handle events from closing and opening the plugin window.

It also seems like you are adding a `localObserver` without actually using it. Not sure why.

Hallbergs (approved):

PR #1271, opened 2023-01-20
Feature/1269 props as filter args

### LINKED ISSUE #1269: InfoClick prop filters - Allow props as filter arguments
There are actually 2 related things that should/could be adressed at the same time.

1. The first thing is that comma (,)  cant be used in an argument because split(',') is used.
Example of what breaks the rendering:  `{propName|fallback('This string will be cut, yes it will.')}`

2. The second thing that would be nice is to be able to use a prop as an argument.
Example: `{propName1|fallback(propName2)|toUpper}`
Triple fallback example (_In theory_) :) : `{propName1|fallback(propName2)|fallback(propName3)|fallback('Hey! You got no value!')}`

jesade-vbg:
closes #1269 

Some other features crept in, but it should not have a negative effect.

### DIFF
--- a/new-client/src/components/FeatureInfo/FeaturePropsFilters.js
+++ b/new-client/src/components/FeatureInfo/FeaturePropsFilters.js
@@ -26,6 +26,13 @@ class PropFilters {
       }
     }
 
+    // Handle different kind of arguments using this RegEx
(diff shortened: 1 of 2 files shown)


jacobwod (approved):
Nice additions!

Hallbergs (approved):
lgtm

PR #1272, opened 2023-01-24
Feature/1252 control hajk programatically

### LINKED ISSUE #1252: Add an API to control certain Hajk features programmatically
_Edit: Scroll down to read on about the methodology I'm thinking of for this issue._
A common use case for Hajk is to be embedded, as an IFRAME, on another website.

This has been known for a while, and the #192 _Clean mode_ has been around ever since.

We now see an increasing demand for allowing programatic control of certain Hajk features, including:
- zoom level (`z` param)
- x and y coordinates (`x` and `y`)
- search query string (`q` param)
- layer visibility (`l` and `gl` params)

(Add more ideas!)

---
_UPDATE 2023-02-08_
**How to enable**
- For now this functionality can only be activated on a per-map-basis using the setting in Admin UI
- If you want to try it out on an entire instance, there's the possibility of adding ?enableAppStateInHash to your query string (e.g. `https://localhost:3000/?enableAppStateInHash`). This will _force_ the functionality on all maps running on this instance, no matter the setting in Admin UI.

**What it does**
- Hajk reacts to changes in hash portion of the URL string
- Hajk keeps the hash updated
- The following parameters work:
  - m (map config)
  - x, y (coordinates)
  - z (zoom level)
  - q (search query)
  - l (layers)
  - gl (sublayers [for group layers]
  - p (visible plugins)

DEMO
This demo also showcases how to embed (and control) Hajk programatically by modifying the hash part of IFRAME's `SRC` attribute.

https://user-images.githubusercontent.com/110222/214314476-4871143f-febf-413a-9450-be9846bfd461.mov

---

I will create a branch where we can work upon this, complete with a nice example with an embedded Hajk:

https://user-images.githubusercontent.com/110222/205056846-9dce8a8a-2a1b-40a8-8046-cba6f815f0dc.mov

---

# Methodology

This feature can be broadly broken up into the following three components:

1. Retrieve app state and **show an updated list of hash params** in URL
2. Ensure that initial app state is **derived** from hash params too
3. Act when any of the hash parameters' value changes

## 1. Retrieve app state and **show an updated list of hash params** in URL
This is quite easy as we already have a mechanism in place that reads out the "relevant" parts of app state and encodes them into a string: we do exactly this in `AnchorModel.js`. So the only thing needed is to update `window.location.hash` when any parameter changes. 

## 2. Ensure that initial app state is **derived** from hash params too
This part is not consolidated, as opposed to part 1. Some parameter values are read in `AppModel.js`, some are read in their respective plugins (e.g. _Search_ looks for values on `q`). And some are on even different places (`index.js` reads out the `m` parameter to know which map should be loaded).

## 3. Act when any of the hash parameters' value changes
This part is, just as part 2, not consolidated. Each layer has a listener where it looks out for its' own `id` in the list of `l` parameter. Search doesn't care about `q` other than for the initial load. And so on.

So we need to set up a lot of listeners to act upon relevant changes. Some will go into `bindHandlers` in `AppModel.js` (perhaps those related to the `Map` state, such as `x`, `y` and `z`. Some will go into layers. Some will go into respective plugins.

jacobwod:
Closes #1252 

https://user-images.githubusercontent.com/110222/214319572-c332ae5e-3e30-4fc0-93d6-89207ce36f86.mov


Hallbergs (dismissed):
Very nice! Works great! 🥇 🔥 🥳  💯

Hallbergs:
Very nice! Works great! 🥇 🔥 🥳  💯

Hallbergs (approved):
Very nice! Works great! 🥇 🔥 🥳 💯

jesade-vbg:
"Snajsigt!", as you say in Swedish. Nice work!

PR #1279, opened 2023-02-03
Feature/1268 accordion blockqoute

### LINKED ISSUE #1268: DocumentHandler - Accordion option for blockqoute
There is a request to add a collapse (accordion) function when users are creating blockqoutes in admin. The purpose is mainly for longer text paragraphs that could be collapsed. 

There should exist a checkbox for toggling if a blockqoute is or is not collapsible:
image

Selecting a paragraph should also give the information if a blockqoute is or is not collapsible:
image

Finally, in print, collapsed blockqoutes should be printed as non-collapsed as a default. The user could choose to print with paragraphs collapsed. I guess the user could choose which collapsed blockqoutes to print and not to print.

OlofSvahnVbg:
closes #1268 

NOTE: 
In the TextAreaInput there is two types of variables that seems to be similar (ex. isAccordion which is a state and currentIsAccordion which is a const). However, the two variables are used to keep track of two different things, the user interaction and the data from the existing blockqoute.

The useState and useEffect I use to keep track of this might not be best practice. If not, I would like to learn of a better way to do it, if that is possible. Thanks.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Stage 1 - Building the backend
-FROM node:16-alpine as backendBuilder
+FROM node:18-alpine as backendBuilder
 WORKDIR /usr/app
 COPY /new-backend/package*.json ./
 RUN npm ci
@@ -7,7 +7,7 @@ COPY ./new-backend .
 RUN npm run compile
 
 # Stage 2 - Building the client
-FROM node:16-alpine as clientBuilder
+FROM node:18-alpine as clientBuilder
 WORKDIR /usr/app
 COPY /new-client/package*.json ./
 RUN npm ci --ignore-scripts
@@ -17,7 +17,7 @@ RUN mv ./public/appConfig.docker.json ./public/appConfig.json
 RUN npm run build --ignore-scripts
 
 # Stage 3 - Building the admin
-FROM node:16-alpine as adminBuilder
+FROM node:18-alpine as adminBuilder
 WORKDIR /usr/app
 COPY /new-admin/package*.json ./
 RUN npm ci --ignore-scripts
@@ -27,7 +27,7 @@ RUN mv ./public/config.docker.json ./public/config.json
 RUN npm run build --ignore-scripts
 
 # Stage 4 - Combine everything and fire it up
-FROM node:14-alpine
+FROM node:18-alpine
 WORKDIR /usr/app
 COPY /new-backend/package*.json ./
 RUN npm ci --production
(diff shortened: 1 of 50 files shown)


Hallbergs:
I'll give this a proper look next week. Regarding the `useEffect` i'm pretty sure you will get an exhaustive-deps warning (since you are referring to some variables without including them in the dependency array).

Hallbergs:
I tried removing all the consts, simplifying the logic quite a bit. However, i have never used the document-handler-editor, so i'm not sure if i ruined something or not 😨 

Everything seems to work with the new logic. Would it be OK if i push it and you test it again? (Sorry for the extra work, but it felt wrong storing everything twice :) ). @OlofSvahnVbg

OlofSvahnVbg:
Of course, go for it! As long as the functionality is the same as before :) @Hallbergs

Hallbergs:
Sweet! Just pushed it. Let me know if it doesn't seem to work as before. If something is wrong i'll revert the "fix". :)

OlofSvahnVbg:
This seems to be working perfectly :) Big thanks for the update. I'm guessing the arrow functions have something to do with not needing the two variables?

Hallbergs:
> This seems to be working perfectly :) Big thanks for the update. I'm guessing the arrow functions have something to do with not needing the two variables?

Kind of. I saw that the constants and the internal state had the same values when the props updated. Therefore the constants could be removed!

Hallbergs (dismissed):

Hallbergs:
I don't know if @jacobwod wants to check the code before merging?

jacobwod:
Hi @OlofSvahnVbg, one question: 
- is this non-destructive, in the sense of "nothing changes with current configs unless you tick the `Make this an accordion` box"?

We have a lot of stuff in production using DocumentHandler and I update frequently from `develop`, so it's crucial that nothing changes to existing documents' look.

OlofSvahnVbg:
I think I made it so that accordion is simply an extra feature. I dont think Ive made changes to existing blockqoutes and that logic. But maybe we can look it over together to be sure, perhaps after todays code meeting?

OlofSvahnVbg:
Or tomorrow works better, actually :)

jacobwod:
Sure, sounds good! 👍

jacobwod:
The last two commits are fine too, merge whenever you're ready.

OlofSvahnVbg:
It seems you need to approve before I can merge @jacobwod :)

jacobwod (approved):
Ooops, missed to tick the `Approve` box, sorry! 😄

PR #1281, opened 2023-02-04
Feature/orebro tsm 2

Sunix71:
A development of the tool document manager 2.0 in new-admin so that you can select multiple events in a story map. And in new-client so that you can click on a document link and get multiple events at the same time. You can now also choose to place an event on a headline menu.


Hallbergs (changes requested):
It seems that the `MenuConnectionSelector` is duplicated? The code is basically the same between all three components. You should create one generic component and re-use that one.

I would suggest reading up on "Thinking in react" or the docs, see: https://beta.reactjs.org/.

Sunix71:
Okay, now I have rewritten the code in one component, but I don't know if it's particularly generic because there are 3 different separate columns being displayed. But with this, I may be one step closer to the goal anyway. Thank you for input !

Hallbergs:
Just added an example of how it can be made more generic. Test it out, and let me know if it works.

Sunix71:
It works very well and is generic in its construction. Very good.

Hallbergs:
Gonna let @jacobwod test with his config, then we can merge

jacobwod:
Tested, seems to work as intended so it's good to go for me.

Hallbergs (approved):

PR #1283, opened 2023-02-06
Feature/1264 bookmarks cookies verification

### LINKED ISSUE #1264: Add verification for cookies usage in Bookmarks plugin
**Description**
The current implementation of the Bookmarks plugin uses localStorage to store bookmarks, but it does not ensure that the user has consented to the use of cookies.

**Suggested Solutions**
- Do not store any data in localStorage if the user declines the use of cookies.
- Inform the user within the plugin if they have declined to use cookies and add a button to enable it again.

linusfj:
Closes #1264.


Hallbergs (approved):
Nice!

PR #1292, opened 2023-02-23
Feature/1243 new measure tool

### LINKED ISSUE #1243: New Measure tool - step 1
I think we should rewrite the measure tool (or probably create a completely new measure tool).
We should also rethink the UI, probably use buttons instead.

There are many other ideas regarding the measure tool in other github issues, but my thoughts is to do this in steps.

**So for step 1 the suggestion is (initial thoughts):**

- Rewrite to functional component
- Use new DrawModel functionality etc
- Use buttons for the different measurement types
- Keep the base functionality simple.

**I think simplicity is key to keeping this tool a success.**

_(When I say success, I base it on the usage statistics of the measure tool)_

Any thoughts?

jesade-vbg:
A long overdue PR.
Closes #1243


Hallbergs (approved):
Looks good! `MeasurerModel` is passed to `MeasurerView` but is not used and can be removed from the props. Maybe the `MeasurerModel` can be removed since it is barely used ;)

Hallbergs (approved):

jacobwod (dismissed):
Fantastic! Just a small addition for macOS compatibility (the Cmd+Z is customary). Works great for me, good job!

jacobwod:
Works fine now regarding the delete interaction.

One last thing I noticed is that the layer is called `Draw layer` internally - a name that's already taken so we end up with duplicates in the Active layers list (please switch on `Show system layers` to see them). 

Is there a chance that this name can be customised and called something more plugin-specific? Perhaps @Hallbergs knows too?

jesade-vbg:
Yes its hardcoded in DrawModel, I will look at it. I have never used the "Active layer" feature so I havnt noticed these things :/

jesade-vbg:
So..... now thats addressed too :)

Hallbergs (approved):

PR #1295, opened 2023-03-02
Vt/backend dotnet pr

sweco-semtto:


Hallbergs (changes requested):
I think there are some major merge mistakes here? Seems like the entire project for the Gothenburg parking company is included? There are several strange changes connected to the VT-plugin as well.

I am expecting a PR on the new backend, nothing more.

Hallbergs:
image
?????

jacobwod:
Unfortunately there are numerous problems with a PR like this. 

First and foremost, please ensure that the PR **only contains the relevant parts**, in this case it's supposed to only contain the new .NET Core backend. 

Also, please provide a description and at least a reference to the relevant issue(s).

Issue #1210, opened 2022-10-11
New .NET-backend


sweco-semtsg:
Update/redevelop .NET-backend to .NET 6 according to the attached OpenAPI-specification.

jacobwod:
Merged in #1395

PR #1297, opened 2023-03-03
New .NET backend

sweco-semtto:
Issue #1210, the new .NET backend.


Hallbergs (changes requested):
## Found issues

Below are descriptions of the issues that were found during the review.

### Base URL

The base url differs from the nodeJS-backend base url. (The new backend lacks /api/v1). The reasoning behind the base url in the nodeJS-backend was that versioning would be simple. For example, if we wanted to add consolidated loading, we could add that under /api/v2 until we were ready to remove the old endpoint.
Maybe the idea is to handle this in the hosting solution?

### Consolidated loading

In the nodeJS-API, a client can access the entire configuration (all layers and map-information) under `/api/v2/config/${map}`. This does not seem to be implemented in the new backend.

### Interchangeable backends

It is not possible to change between the backends in a simple manner (mainly because of the base-url problem mentioned above). This leads to unnecessary headache both when developing, but also when setting up docker-images etc. In the configuration files for the docker images, we are targeting the backend with `/api/v1` - this won't work with the new .NET-backend, and separate configuration files has to be made for each backend. Something like `appConfig.nodejs.docker.json` and `appConfig.dotnet.docker.json` - seems unnecessary.

### Logging

We're lacking proper logging. It is hard to use the logs to find the reason for a problem. An example: If a user tries to access a map that they are not allowed to, we throw the following: `System.Exception: [getMapConfig] Access to that map not allowed for this user.`. This is later added to the log in a catch-block. Question: Is it really ideal to log `that map` and `this user`? It doesn't provide the admin with any useful information at all.

### Active directory user search

In the nodeJS backend, we use `sAMAccountName` when searching for a user. In the new .NET backend, `userPrincipalName` is used. (In our AD `sAMAccountName` is the username we use to login, in my case henhal2002 - and `userPrincipalName` is the email, in my case henrik.hallberg@exploatering.goteborg.se). Using `userPrincipalName` might be correct nowadays, but since it differs from the nodeJS backend i suggest that we add this as a setting.

### Security concerns

It doesn't seem to be any IP-restrictions when grabbing `userPrincipalName` from the headers. This means that anyone could send requests to the new backend as an admin user and access all endpoints. **Is the idea that this should be handled in the hosting solution? If so, it should be added to the readme.**

To be clear: If someone was to use this in production and deploy according to your readme, i could possibly perform a `DELETE` against `https://hajk-public.some.municipality.se/mapconfig/${map}` and succeed if i know the email of an HAJK-admin in that municipality.

### Hard coded header-key for LDAP search

To enable the AD functionality the user-information has to be sent in the `X-Control-Header` key. Maybe this could be configurable?

### LDAP connection string

To get the LDAP connection working, i had to provide the connection string as `LDAP/...` instead of `ldap/...` as stated in the readme. (In the nodeJS-backend i can use lowercase so this is specific to the new backend). Please make sure the instructions in the readme is correct.

### Using the `visibleForGroups` property

To be able to filter configuration objects inside a map-config, the entire map-config must have restrictions in the provided solution. For example, it is not possible to hide a layer in a map that should be visible to everyone. It is common that a map is un-restricted while having some tools etc. restricted.

### Tools not responding to the `visibleForGroups` property

If i set a tool to be visible only for some groups, it is not removed for clients not in that group. I think the entire tool-filtering is missing.

### `visibleForGroups` logic error

When using the AD-restrictions, you make a check to see if the user has access to the map in the following way:
`if (visibleForGroups == null) { return filteredMapObjects; }`
In the configuration you provided, `visibleForGroups` is set to an array (as it should) which means the entire map-config is always returned. A better check would be:
`if (visibleForGroups == null || visibleForGroups.Count() == 0) { return filteredMapObjects; }`

### `config/${map}` marked as obsolete

Why is this endpoint marked as obsolete? How is client supposed to retrieve the map config?

### Unused key in `appSettings`

Is `IdentifyUserWithWindowsAuthenticationX` used? If yes, what does it do? I cannot find any reference to it in the code.

### Unclear naming of `appSettings` keys

The groups that should have access to admin-endpoints are to be added in the `groups`-key in `appSettings`. `Groups` could be given a name that is easier to understand.

### Missing CORS support in development mode

I cannot find any CORS-settings - making it cumbersome to develop against the new .NET-backend. (Since the client and backend are on different hosts when developing).

## Conclusion

The .NET-backend in the current PR **is not in such a state were it can be added to Hajk**. Before considering adding it, the issues above must be addressed in some way.

Hallbergs:
Publish new PR when ready!

PR #1303, opened 2023-03-24
Fix/1290 checkbox indicator group layers

### LINKED ISSUE #1290: Min/max zoom level - Checkboxes do not turn red for group layers when outside zoom level
**Bug deskription for group layers**
When you zoom outside a specified zoom level and switch on a single layer, the checkbox in the layer manager turns red. When you are outside a specified zoom level and switch on a group layer, the checkbox in the layer manager does not turn red.

**Expected behavior and wishes for groups**
That the indicator (red color in checkboxes) that shows that you have zoomed outside a specified zoom level, work the same on group layers as for individual layers. So that red checkboxes is displayed when the layer is lit outside the zoom level.

This is a clarification of the bugs described in #810

linusfj:
Closes #1290.


jacobwod (changes requested):
Unfortunately it does not work. In short: state does not refresh when needed and the indicator changes color only under some very specific conditions. See attached video.


https://user-images.githubusercontent.com/110222/227876866-e01932b7-043b-47cd-a53d-6c86fa0a0077.mov

Issue #1300, opened 2023-03-20
Min/max zoom level - Too many snackbars are displayed


elizabethgbg:
Only one snackbar should be displayed when one or more layers are switched on outside layers specified zoom level (or when zooming outside layers zoom level and have one or more layers switched on, depending on the admin setting). This applies to groups, layers and layer groups.

Text suggestions for the snackbars:
For one layer: "Lagret *lagernamn* är inte synligt vid aktuell zoomnivå."
For multiple layers (for groups, layers and layer groups): ”Lagret *lagernamn* och xx andra lager är inte synliga vid aktuell zoomnivå.”

jacobwod:
Hi! Although a valid issue description, I get a feeling that we are creating duplicates of existing issues. As far as I can tell, all except one in the screenshot below are related:
![Skärmavbild 2023-03-22 kl  08 06 01](https://user-images.githubusercontent.com/110222/226826986-3ec53342-9285-4da7-9adb-809846424cf0.png)

It'd be great if you or someone in your organization with admin permissions on GitHub could consolidate related issues and close some of the duplicates. Thank you!

Issue #1284, opened 2023-02-06
Add theme management


linusfj:
### Description
`Quick Layers` is a feature to manage layers within the `LayerSwitcher` tab. The purpose of the feature is to get quick access to layers often used. One advantage of using `Quick Layers` is that a collection of layers can be included where not all of them need to be visible from the start. Layers can be added manually or loaded from pre-defined packages for all users. The package owner creates the `Quick Layers`, exports the settings to JSON, and the administrator registers the new package of `Quick Layers`. It will also be possible for the user to save, edit and load their own `Quick Layers`.

---

### Proposed Features

- `Quick Layers` Group:
  - Add a new group called "Quick layers" at the top of the layers list in the `LayerSwitcher`.
  - Include buttons to load pre-defined and user-defined collections of layers and delete all layers from the `Quick Layer` group.

- Functionality:
  - Ensure layers within the `Quick Layer` group behave similarly to other layers in the `LayerSwitcher` (visibility toggle, access layer information).
  - Add an option to add regular layers outside the `Quick Layer` group to the `Quick Layer` group.

- `Quick Layer` Package Components:
  - Title, Author, Description, Keywords, Layers, Visibility, Drawing Order, and Settings.

![image](https://user-images.githubusercontent.com/8030104/228553353-099a6883-4c54-445b-a5bc-a2b5be1a8c31.png)

---

### Loading pre-defined `Quick Layers`

1. Click the “Open” button in the `LayerSwitcher` to load pre-defined collections of layers.
2. Choose a pre-defined `Quick Layer` package from the list.
3. The selected `Quick Layer` package will load, replacing the current layers and settings in the `LayerSwitcher` tab.

![image](https://user-images.githubusercontent.com/8030104/228553470-e5dcc8f8-de6c-4d68-ad72-a1bbbedc46e9.png)

---

### Creating user-defined `Quick Layer`

1. In the `LayerSwitcher`, press the “More” icon and click on the option “Add to Quick layers” to build a collection of layers.
2. Click the “Person” button in the `Quick Layers` Group for personal packages.
3. Enter a Title, Owner, and Description for the new `Quick Layer` collection.
4. Click "Save" to create the user-defined `Quick Layer` collection.
5. The newly created `Quick Layer` collection will be stored using `Local Storage` and is available to load.

![image](https://user-images.githubusercontent.com/8030104/228553967-0fd69d16-bfd3-4c8b-9e91-27032adaacd4.png)

---

### Example of JSON structure

[...]

PR #1304, opened 2023-03-24
Fix/1291 snackbar group layers

### LINKED ISSUE #1290: Min/max zoom level - Checkboxes do not turn red for group layers when outside zoom level

### LINKED ISSUE #1291: Min/max zoom level - Snackbar not displayed for layer groups when switched on outside of zoom level
**Bug deskription for group layers**
When you zoom outside of a layers specified zoom level and switch on a normal layer, a snack bar will appear.
When you zoom outside of a layers specified zoom level and switch on a group layer, no snack bar will appear.

**Expected behavior and wishes for group layers**
That the indicators for the layer being turned on outside the specified zoom level, work the same on group layers as for single layers. So that the selected setting for the snackbar is displayed for both group and single layers.

**NOTE**
There are settings in admin for the snackbar - the setting determines if the snackar only should be displayed when the layer is switched on outside the specified zoom lavel, or every time you zoom outside the zoom level when the layer is switched on. Group layers should work like single layers both when this function is activated and deactivated.

This is a clarification of the bugs described in #810 
Related issues: #1290, #1299, #1300.

linusfj:
Closes #1290 and #1291.


new-client/src/plugins/LayerSwitcher/components/LayerGroupItem.js
```diff
@@ -582,6 +624,26 @@ class LayerGroupItem extends Component {
     );
   };
 
+  showZoomSnack(subLayer) {
+    if (this.zoomWarningSnack) return;
```
jacobwod, line 628: Can you expand with some relevant comments? Here is a good place as it might at first seem counter-intuitive (why are we just leaving the method if `zoomWarningSnack` is truthy?).

jacobwod (commented):
Which of the proposed solution does this PR implement? We had a couple of ideas, one of them being the "Layer X and n other layers can't be shown". Is it this, or is it more like the other solution in `LayerItem`? 

Btw, does this play well with `LayerItem`'s solution, or will we only get more and more snackbars stacked on each other?

new-client/src/plugins/LayerSwitcher/components/LayerGroupItem.js
linusfj: The method ensures that only one `Snackbar` notification is displayed at a time, preventing multiple notifications from overlapping.

linusfj:
> Which of the proposed solution does this PR implement? We had a couple of ideas, one of them being the "Layer X and n other layers can't be shown". Is it this, or is it more like the other solution in `LayerItem`?
> 
> Btw, does this play well with `LayerItem`'s solution, or will we only get more and more snackbars stacked on each other?

In this solution, the messages will stack as they do for `LayerItem`. A suggestion to combine them into one `Snackbar` is described in issue #1300.

jacobwod:
There is some improvement here as we do get the sublayer's name to show. Unfortunately, the behaviour is still far from good UX in my opinion, at least on this setup. I've attached a short video so that you can see what I mean.

My understanding was that this issue would limit the amount of sub-sequential snackbars being display and that's unfortunately not the case as you can see:

https://user-images.githubusercontent.com/110222/234532474-30bb7b9b-9bb9-4068-95c1-a169c2e51011.mov

jacobwod (changes requested):
In addition to the latest video attached, where the functionality does not seem optimal, I have two small issues with this:
- Can you please add some _comments in the code_ as the flow starts getting complex. I see you added a comment to my comment here, but that wasn't what I meant and I'd truly appreciate some comments that explain your thoughts behind the implementation. E.g. compare the amount of comments in the portion above the recently added:
Skärmavbild 2023-04-26 kl  11 33 27
- There's a linter warning about unused variable

Just to be clear: the picture above is just an example, comments are lacking in most places in the new code.

linusfj:
Yes, one of the initial ideas was to limit the number of `Snackbar` messages when turning on a group and include it in this Pull Request, but it has instead been added as issue #1300.

There's also ongoing work in parallel with #1284, where part of the effort is rewriting the layer components. The plan is to continue the development for limiting `Snackbar` messages (#1300) built on top of that code and develop it against a copy of the "1284-add-theme-management" branch. Since the `LayerSwitcher` project is still in progress, those changes will be included in the code delivery after the project has been completed.

Additionally, issues #1290 and #1291 will eventually be adapted for the new layer components in that branch.

In summary, this Pull Request aims to update layer groups (`LayerGroupItem`) to have the same functionality as regular layers (`LayerItem`), despite the presence of some apparent suggestions for improvement.

jacobwod (approved):
If you've confirmed that:
  - this works both for raster and vector data sources (Hajk's `vectorlayers`)
  - the functionality is as desired
  - there are no compile/lint warnings
  - and **you've successfully merged the _latest_ `develop`**

…then, feel free to merge!

PR #1315, opened 2023-04-05
Fix that closes #1314 -dh2.0-large-images-breaks-view

### LINKED ISSUE #1314: DocumentHandler 2.0 - Large images with width and height set breaks view.
Today, both width and height is set. If an image is too large... the view wont look good.

Sollution:
Only width should be set, then height will be calculated by the browser.
And the fail safe should be to add maxWidth: 100%. This is the common way to handle it using css.

jesade-vbg:
Closes #1314

### DIFF
--- a/new-client/src/plugins/DocumentHandler/utils/ContentComponentFactory.js
+++ b/new-client/src/plugins/DocumentHandler/utils/ContentComponentFactory.js
@@ -442,7 +442,7 @@ export const Img = ({ imgTag, localObserver, componentId, baseUrl }) => {
         sx={{
           ...getImageStyle(image),
           ...(image.width && {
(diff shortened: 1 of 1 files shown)


Hallbergs (approved):
In the examples (`exempeldokument.json` etc.) both width and height are set, and the solution seems to work great! 

Is this going to be a breaking change? I.e. are there users with only height set? Maybe the img-refs are always saved with both height and width from admin? In that case we should be safe!

jesade-vbg:
> In the examples (`exempeldokument.json` etc.) both width and height are set, and the solution seems to work great!
> 
> Is this going to be a breaking change? I.e. are there users with only height set? Maybe the img-refs are always saved with both height and width from admin? In that case we should be safe!

Yes, admin only accepts empty width and height or both filled. Should not break anything.

PR #1318, opened 2023-04-11
feature/1309: API version separation in NodeJS backend

jacobwod:
Pretty massive but should work.

Some new keys in .env can be used to customise the functionality, but **my primary intention is not to require any new keys, so please test with your current `.env`**.


Hallbergs (approved):
Beautiful!

PR #1322, opened 2023-04-14
Feature/PrintAttachmentsMerge

### LINKED ISSUE #1289: DocumentHandler - Print attachments
We want to implement a way to print attachments in the Print menu of Document 2.0:

image

The idea is to create a submenu for printing attachments below the current document print menu. We might need to extend the scrollbar to encapsulate both the checkboxes and the print button. Otherwise we're left with two scroll bars, the current one and the one that will be needed for scrolling past the document print menu.

Exactly how we will display the choice for printing which attachments is not yet determined. However, a new menu like the existing document print menu might be a good solution.

OlofSvahnVbg:
This is a branch created solely for merging the relevant changes made in #1289 (DocumentHandler - PrintAttachments). There was some code overlap with another branch that has already been merged.

closes #1289

### DIFF
--- a/new-client/src/plugins/DocumentHandler/printMenu/PrintWindow.js
+++ b/new-client/src/plugins/DocumentHandler/printMenu/PrintWindow.js
@@ -62,7 +62,6 @@ const GridMiddleContainer = styled(Grid)(({ theme }) => ({
     },
   },
   flexBasis: "100%",
-  maxHeight: "400px",
 }));
 
 const GridHeaderContainer = styled(Grid)(({ theme }) => ({
@@ -74,7 +73,6 @@ const GridSettingsContainer = styled(Grid)(({ theme }) => ({
 }));
 
 const GridFooterContainer = styled(Grid)(({ theme }) => ({
-  marginTop: "10px",
   flexBasis: "10%",
 }));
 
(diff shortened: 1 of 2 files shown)


new-client/src/plugins/DocumentHandler/printMenu/PrintWindow.js
```diff
@@ -840,30 +903,19 @@ class PrintWindow extends React.PureComponent {
     );
   };
 
-  render() {
-    const { togglePrintWindow, localObserver, documentWindowMaximized } =
-      this.props;
+  toggleDocumentsAttachments = () => {
+    let showAttachments = this.state.showAttachments;
+    this.setState({ showAttachments: !showAttachments });
+  };
```
Hallbergs, line 909: This could lead to some state-timing issues, i would suggest:

```js
this.setState((prevState) => ({ showAttachments: !prevState.showAttachments }));

Hallbergs (changes requested):
Remove warnings please:
image
Seems like there are two identical functions!

I see that there are some style-changes to the print-window. Is this print-button move desired? (Moving the button from the bottom to the middle-ish).
Before (In 3.11 at least, maybe this was introduced somewhere else?):
image

In the PR:
image

new-client/src/plugins/DocumentHandler/printMenu/PrintWindow.js
OlofSvahnVbg: Fixed!

Hallbergs (approved):

PR #1324, opened 2023-04-14
Reinstated cacheBuster to new-client

### LINKED ISSUE #1323: hfetch - reinstate the cacheBuster functionality.
The cacheBuster fuctionality was disabled little over a year ago because we updated to webpack 5 which does not have support for process.env.

The cacheBuster fuctionality in hfetch should be reinstated again.... somehow.... with or without process.env.

Today, the code is just commented, and forgotten.

`this.useCacheBuster = process?.env?.REACT_APP_USE_CACHE_BUSTER === "true" || false;`

jesade-vbg:
Closes #1323

### DIFF
--- a/new-client/src/utils/FetchWrapper.js
+++ b/new-client/src/utils/FetchWrapper.js
@@ -43,11 +43,11 @@ class FetchWrapper {
     const appName =
       document
         .querySelector(`meta[name='app-name']`)
-        ?.getAttribute("content") ?? "";
+        ?.getAttribute("content") || "";
     this.hash =
       document
         .querySelector(`meta[name='${appName}-git-hash']`)
-        ?.getAttribute("content") ?? "";
+        ?.getAttribute("content") || "";
     this.useCacheBuster =
       document
         .querySelector(`meta[name='${appName}-use-cache-buster']`)
--- a/new-client/src/utils/FetchWrapper.js
+++ b/new-client/src/utils/FetchWrapper.js
@@ -39,10 +39,12 @@ class FetchWrapper {
     this.options = {};
 
     // Hash is used for cacheBuster function.
-    //this.hash = process?.env?.REACT_APP_GIT_HASH || null;
-    this.hash = null;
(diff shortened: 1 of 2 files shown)


new-client/src/utils/FetchWrapper.js
```diff
@@ -57,6 +59,14 @@ class FetchWrapper {
...
   }
 
+  getMetaValue(key) {
+    const el = document.getElementsByName(key);
+    if (el && el[0]) {
+      return el[0]?.attributes?.content?.value;
+    }
+    return null;
+  }
+
```
jacobwod, line 69: Is this really necessary, wouldn't something like this be more sufficient and work on all modern browsers anyway?
```javascript
document.querySelector("meta[name='hajk-client-use-cache-buster']").getAttribute("content");
```

new-client/src/utils/FetchWrapper.js
```diff
@@ -39,10 +39,12 @@ class FetchWrapper {
     this.options = {};
 
     // Hash is used for cacheBuster function.
-    //this.hash = process?.env?.REACT_APP_GIT_HASH || null;
-    this.hash = null;
-    //this.useCacheBuster = process?.env?.REACT_APP_USE_CACHE_BUSTER === "true" || false;
-    this.useCacheBuster = false;
+    // Lets get the values from generated meta-tags
+    this.hash = this.getMetaValue("hajk-client-git-hash") || "";
```
jacobwod, line 43: Can we be sure that the meta tag will always be called `hajk-client-use-cache-buster`? I mean, it's created during build time by parsing `%REACT_APP_NAME%-use-cache-buster`, so another value on `REACT_APP_NAME` will break this completely.

jesade-vbg:
You're absolutely right regarding these 😂 I should never make PR:s on Friday afternoons 😂  I'll fix on Monday.

jesade-vbg:
Now it's corrected. Thanx

jacobwod (approved):
Looks good to me! 👍

PR #1327, opened 2023-04-19
Getting custom fonts to work in DocumentHandler Print.

jesade-vbg:
First tests seems promising but I'll test some more on other servers before making it a real PR.


jesade-vbg:
Now we've tested with good results!

Hallbergs (approved):

Issue #1328, opened 2023-04-20
Ensure that NodeJS backend is fully ESM-compatible


jacobwod:
While an attempted routine dependency upgrade I run into some problems: I couldn't upgrade to `query-string` >=8. I got some unexpected errors and it seems to have to do with the way we import modules (EMS vs CommonJS). This got me baffled, as I thought that the new Backend was using ESM since start (after all there are no `require()` and we `import` all our packages). 

But it turns out that there are some more steps which need to be taken into account. A nice list I [found here](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c) says this:

> Add "type": "module" to your package.json.
Replace "main": "index.js" with "exports": "./index.js" in your package.json.
Update the "engines" field in package.json to Node.js 14: "node": ">=14.16". (Excluding Node.js 12 as it's no longer supported)
Remove 'use strict'; from all JavaScript files.
Replace all require()/module.export with import/export.
Use only full relative file paths for imports: import x from '.'; → import x from './index.js';.
If you have a TypeScript type definition (for example, index.d.ts), update it to use ESM imports/exports.
Optional but recommended, use the [node: protocol](https://nodejs.org/api/esm.html#esm_node_imports) for imports.

So I'm creating this issue to get this on my radar and I'll attempt to fix it in the near future.

jacobwod:
I'm working on cleaning up some (old) rules and plugins. One commit is upcoming that will remove Babel, which - of course - is a big step.

In the meantime, some checks I'll be comparing against. I'm half-way through this change and both ESlint and Prettier are good so far:
![Skärmavbild 2023-04-21 kl  11 23 09](https://user-images.githubusercontent.com/110222/233599628-63fe26a3-37af-4e3e-94cf-0e65e765af31.png)

jacobwod:
One small issue remains: as I expected, the lack of `__dirname` when running as ESM and the change I made in 83e9d66bb61d71b21203db6efe2267122e872369 leads to this when running in Docker:

![Skärmavbild 2023-04-21 kl  14 40 46](https://user-images.githubusercontent.com/110222/233638545-bf6c08ab-0f2d-4670-b3d2-6d17aeed479c.png)

Not shocking at all, as the image looks like this:
![Skärmavbild 2023-04-21 kl  14 41 25](https://user-images.githubusercontent.com/110222/233638679-ac19c1bc-fa71-47d8-a69c-9afa76fd4d1e.png)

[...]

PR #1332, opened 2023-04-24
Make the Backend fully ESM compatible, remove Babel

jacobwod:
This PR closes the issue described in #1328. Originally, I inteded this only to fully migrate the codebase to ESM, but it turned out there's no reason to transpile by now (as anybody using the latest version should be on Node's LTS anyway, while we support one major version backwards with this release, i.e. NodeJS 16.10). 

This is a big change however and it could potentionally break someone's workflow. So please CHECK IT OUT AND TEST LOCALLY and perhaps in some non-local environment. 

Perhaps the most dramatic changes are to be found in package.json and the various .*.rc files. Make sure to see what's been changed. 

I did the tests a) locally, b) by building it out to a directory and c) by building a Docker image. It seems fine so far but more organizations should try this out. Thanks.

### DIFF
--- a/new-backend/.prettierrc
+++ b/new-backend/.prettierrc
@@ -5,5 +5,5 @@
   "trailingComma": "es5",
   "bracketSpacing": true,
   "printWidth": 80,
-  "endOfLine": "auto"
+  "endOfLine": "lf"
 }


Hallbergs (commented):
Very nice! 🥳 

I've tested it on the following without issues:
- Windows 10
- Docker (Ran the image on Windows 10)

However, when deploying to Windows server using IISNode, i ran into some issues...
IISNode is using `require` to run Hajk's `index.js`, resulting in the following error:
image

It is solved by changing a line in one of IISNode's files (`interceptor.js`). It's a small change, see below:
image

I guess that as long as we document this, it should be OK? I'm running transpiled applications on the same server, using the same IISNode without issues even after the change.

PR #1337, opened 2023-05-12
Potential fix for #1320:

### LINKED ISSUE #1320: Getting map config for restricted map can sometimes fail and return `[ERROR] service.auth.v2 - read ECONNRESET`
**Describe the bug**
When AD connection is active and a map config is limited to certain groups, Backend will to a check by grabbing current user's groups from the AD. This request seems to occasionally crash, leading to an empty array being returned, hence Backend thinks that user doesn't belong to any group and the access to map is rejected.

At first I thought it was related to the recent changes in API versioning. **But I can confirm the same behaviour in the commit before PR merge, i.e. this error happens not only in latest `develop` but also in 2759c5574dd1748b88ca47fc8849272450316503 and probably the commits before that.**

**To Reproduce**

1. Checkout `develop`
2. Make sure your `.env` is configured to use AD authentication. In addition, mine uses an LDAPS connection and I suspect this has to do with the certificates somehow, so if you can, try using an LDAPS as well.
3. Make sure that you have a map config set up and limit the access to that map to some AD group that the user you will be testing with belongs to.
4. Fire up the Backend. Make sure that you see something like this in the log:
```
[2023-04-13T14:35:40.486] [TRACE] service.auth.v2 - Initiating ActiveDirectoryService V2
[2023-04-13T14:35:40.487] [TRACE] service.auth.v2 - Setting up AD connection to: ldaps://<domain>
[2023-04-13T14:35:40.489] [INFO] service.auth.v2 - Testing the AD connection to ldaps://<domain>…
[2023-04-13T14:35:40.557] [TRACE] service.auth - Initiating ActiveDirectoryService
[2023-04-13T14:35:40.557] [TRACE] service.auth - Setting up AD connection to: ldaps://<domain>
[2023-04-13T14:35:40.558] [INFO] service.auth - Testing the AD connection to ldaps://<domain>…
```
5. Request the map config from the V2 API. You can do it in your browser, or `curl` or whatever you prefer. Basically, `GET http://localhost:3002/api/v2/config/{yourMapConfig}`
6. If you can see the valid output, you're fine. In that case **restart the Backend** (by typing `rs` and hitting `Enter` in the console where Backend is running). 
7. **But if you got a message saying that `Error: [getMapConfig] Access to map "{yourMapConfig}" not allowed for user "{yourUserName}"`, you hit the same error as I do.** From time to time… Which is really difficult to test.
8. So, if you had the error above, I'm certain that you can see this in your log: 
```
[2023-04-13T14:38:35.243] [TRACE] service.auth.v2 - [getGroupMembershipForUser] No entry for '{yourUserName}' in the groups-per-users store yet. Populating…
[2023-04-13T14:38:35.935] [ERROR] service.auth.v2 - read ECONNRESET
```

**Additional context**
I'm on NodeJS v18.15.0.

Locally, I logged some more info and got this in return, which makes me think this has to do with the LDAPS part:

```
at TLSWrap.onStreamRead (node:internal/stream_base_commons:217:20) {
  errno: -54,
  code: 'ECONNRESET',
  syscall: 'read'
```

jacobwod:
- See discussion in #1320 
- I made all the new options configurable in `.env`. This should allow us to have unchanged production environments and only configure the desired parts in `.env`.
- The new options in `.env` default to everything (new from this issue) turned off, except for the reconnect flag being set to `true`. I think it's a sane default for most situations.

Closes #1320

### DIFF
--- a/new-backend/.env
+++ b/new-backend/.env
@@ -155,6 +155,15 @@ AD_PASSWORD=pass
 # Leave empty to send an empty query (should work for all cases). 
 #AD_CHECK_CONNECTION_QUERY=
 
+# The followig options allow for better control of reconnection and timeouts.
+# It is recommended to set AD_RECCONNECT to `true`. The various timeout options
+# allow for further customization to ensure a stable LDAP connection, but they
+# are disabled by default. See also http://ldapjs.org/client.html
+AD_RECONNECT=true
+#AD_CONNECTTIMEOUT=60000
+#AD_IDLETIMEOUT=60000
+#AD_TIMEOUT=120000
+
 # Controls whether the AD user object will be sent in the GET map config response.
 # This is a useful way to let Client UI know more about the user and allows for
 # user-specific customization.
--- a/new-backend/server/apis/v1/services/activedirectory.service.js
+++ b/new-backend/server/apis/v1/services/activedirectory.service.js
@@ -98,6 +98,19 @@ class ActiveDirectoryService {
       );
     }
 
+    // Grab some more options from .env regarding auto-reconnect and timeout
+    // and transform to valid config constants. See also: http://ldapjs.org/client.html
+    // and #1320.
+
+    // Try to reconnect when the connection gets lost (Default is false)
+    const optReconnect = Boolean(process.env.AD_RECONNECT);
+    // Milliseconds client should wait before timing out on TCP connections (Default: OS default)
+    const optConnectTimeout = Number.parseInt(process.env.AD_CONNECTTIMEOUT); // ms
+    // Milliseconds after last activity before client emits idle event
+    const optIdleTimeout = Number.parseInt(process.env.AD_IDLETIMEOUT); // ms
+    // Milliseconds client should let operations live for before timing out (Default: Infinity)
+    const optTimeout = Number.parseInt(process.env.AD_TIMEOUT);
+
     // Now we have the AD options and - optionally - the TLS options.
     // We're ready to prepare the config object for AD.
     const config = {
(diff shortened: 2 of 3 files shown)


sweco-semara (approved):
Sensible defaults. New env template contains the most reasonable/recommended settings for LDAPS in production, but backwards compatible if using a current env = expect no impact on any existing working set-up. But the added potential improvement for others, when using the new env template. Useful with log tracing of options, but no security risk due to sensible information leakage since the output is obfuscated first. Looks very good! 👍

jacobwod:
Thanks @sweco-semara, I'll merge this into `develop`. 

Regarding the problems you're running into @ingvar-uddevalla: are you/your contractors able to create a new build using `develop`, or do you have to wait until the next release to try this out? 

(We could, optionally try to do a hot fix to `master` too but I won't be able to test it as we have moved on and our production environment requires some new stuff that only exist in `develop`. That would require someone else's expertise with testing and you still need to be able to create a build from a branch manually, so perhaps there's little to win. But it's an option, should you prefer that.)

ingvar-uddevalla:
Is this "new" .env file available in the Hajk 3.12 rc2 package? And does this mean that the problems with LDAPS in issue #1320 could be solved? (I can see that this issue #1337 has cooments from May, i.e before we really begun to discuss these problems). However, it would be good to know because we're about to "go sharp" in a month and cannot use the uncrypted "LDAP-solution" then. Our IT-department will require LDAPS. Otherwise we have to reduce the information (i.e not show secret data) or let our IIS handle the securitygroups in some way.

jacobwod:
@ingvar-uddevalla I'm not sure what's in specific package, but you are probably using a highly customized .env at this point anyway. So the simples way would be adding the new keys to your .env. See changed files for a diff that clearly shows what's added: https://github.com/hajkmap/Hajk/pull/1337/files.

PR #1339, opened 2023-05-26
Feature/1338 anchor bug fix

### LINKED ISSUE #1338: Oversiktsplan - Url link missing from Anchor View (Dela)
The URL-link is missing from the text window in Anchor View (Dela-verktyg) when loading the page:

image

This can't be reproduced for those without access to oversiktsplan-utv, but it happens when you go to oversiktsplan-utv
Click 
image
And the text window will appear empty.

To get the URL-link to appear, the user needs to click the radio buttons...
image
a few times to make it appear.

The link should be there when the anchor view appears when clicked.

I'm not sure, but the problem seems to be that the getAnchor()-method "uses" debounce (which I do not understands fully) and that is keeping the link from loading correctly when the component mounts. 
If anyone knows more, please advise :)

image

image

OlofSvahnVbg:
This PR refers to this issue: Oversiktsplan - Url link missing from Anchor View (Dela) #1338 

closes #1338

### DIFF
--- a/new-client/src/plugins/Anchor/AnchorView.js
+++ b/new-client/src/plugins/Anchor/AnchorView.js
@@ -39,22 +39,18 @@ class AnchorView extends React.PureComponent {
     cleanUrl: false,
   };
 
-  componentDidMount() {
-    // We check if anchor is a string, to avoid error in render() below.
-    // This because the getAnchor() sometimes returns a Promise instead of a string.
-    typeof anchor === "string" &&
-      this.setState({ anchor: this.props.model.getAnchor() });
-
+  async componentDidMount() {
     // Subscribe to changes to anchor URL caused by other components. This ensure
     // that we have a live update of the anchor whether user does anything in the map.
-    this.props.globalObserver.subscribe(
-      "core.mapUpdated",
-      ({ url, source }) => {
-        this.setState({
-          anchor: this.appendCleanModeIfActive(url),
-        });
-      }
-    );
+    this.props.globalObserver.subscribe("core.mapUpdated", ({ url }) => {
+      this.setState({
+        anchor: this.appendCleanModeIfActive(url),
+      });
+    });
+
+    // Initiate the anchor-url on mount
+    const a = await this.props.model.getAnchor();
+    this.setState({ anchor: a });
   }
 
   appendCleanModeIfActive = (url) =>
--- a/new-client/src/plugins/Anchor/AnchorView.js
+++ b/new-client/src/plugins/Anchor/AnchorView.js
@@ -41,18 +41,9 @@ class AnchorView extends React.PureComponent {
 
   componentDidMount() {
     // We check if anchor is a string, to avoid error in render() below.
-    // This because the getAnchor() sometimes returns a Promise instead,
-    // of a string.
+    // This because the getAnchor() sometimes returns a Promise instead of a string.
     typeof anchor === "string" &&
       this.setState({ anchor: this.props.model.getAnchor() });
-
-    // const anchor = this.props.model.getAnchor();
-
-    // if (typeof anchor === "string") {
-    //   this.setState({
-    //     anchor: anchor,
-    //   });
-    // }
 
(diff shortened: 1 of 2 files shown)


jacobwod:
Can you please check if the behaviour is still correct when `enableAppStateInHash` is enabled? To enable it, just add the query string param `enableAppStateInHash`, e.g. https://karta.halmstad.se/?enableAppStateInHash. 

Do the live hash update _and_ the Anchor plugin (both its initial value _and_ live updating of the input in `AnchorView`) still work?

OlofSvahnVbg:
So I tried https://oversiktsplan-utv.varberg.se/?/enableAppStateInHash and the anchor gets set the same way  as it does without the query, and the live update seems to work aswell (if you mean registering changes like moving the map, zooming etc.)

jacobwod:
That sound good, thanks for the test. We'll take a proper look next week!

new-client/src/plugins/Anchor/AnchorView.js
```diff
@@ -39,11 +39,11 @@ class AnchorView extends React.PureComponent {
...
 
-  async componentDidMount() {
-    // Set anchor URL initially
-    this.setState({
-      anchor: await this.props.model.getAnchor(),
-    });
+  componentDidMount() {
+    // We check if anchor is a string, to avoid error in render() below.
+    // This because the getAnchor() sometimes returns a Promise instead of a string.
+    typeof anchor === "string" &&
```
Hallbergs, line 45: `anchor` is not defined here and won't be of type string. When i test your implementation, the view results in an empty field: 
image

(It renders an empty field if you have the anchor plugin enabled at start).


I think this issue could be resolved with:

```js
async componentDidMount() {
    const a = await this.props.model.getAnchor();
    this.setState({ anchor: a });
    // Subscribe to changes to anchor URL caused by other components. This ensure
    // that we have a live update of the anchor whether user does anything in the map.
    this.props.globalObserver.subscribe(
      "core.mapUpdated",
      ({ url, source }) => {
        this.setState({
          anchor: this.appendCleanModeIfActive(url),
        });
      }
    );
  }
```

Maybe you could test that and see if it works for you as well? The strange thing is that the original implementation also works for me...

Hallbergs (changes requested):

new-client/src/plugins/Anchor/AnchorView.js
OlofSvahnVbg: Interesting. I tried having the dela-window opened at start aswell, and the link does appear. But I'll try your fix :)

OlofSvahnVbg: So your codes seems to be working fine. Although, if I scroll while the page is starting the link does not appear.

jacobwod: You mean it doesn't get updated?

The `componentDidMount` part runs only on initial component load. The following updates of anchor string are taken care of elsewhere, please ensure to use the same methodology (`await this.props.model.getAnchor()`) before calling `setState()`.

Hallbergs: Hmm... Seems like `getAnchor` never resolves in some cases... It's not pretty but this should do the trick: 

```js
async componentDidMount() {
    // Subscribe to changes to anchor URL caused by other components. This ensure
    // that we have a live update of the anchor whether user does anything in the map.
    this.props.globalObserver.subscribe(
      "core.mapUpdated",
      ({ url, source }) => {
        this.setState({
          anchor: this.appendCleanModeIfActive(url),
        });
      }
    );
    const a = await this.props.model.getAnchor();
    this.setState({ anchor: a });
  }
```

(We have to make sure the subscription is made. If `getAnchor` never resolves, the subscription is never made, and the anchor-url is never updated).

OlofSvahnVbg: This works on my end :)

Hallbergs (approved):

jacobwod (approved):

PR #1353, opened 2023-06-21
PR for #1348-Infoclick on invisible layers

### LINKED ISSUE #1348: Infoclick on invisible layers
Infoclick works even when a layer is turned off.

If you turn on a layer that is only visible on certain scales the infoclick still works on the scales where the layer is invisible. 
Se video below:
https://github.com/hajkmap/Hajk/assets/111860819/1a5e04ce-bce5-429a-a37f-29d6fac58f3d

To reproduce:
1. Go to https://karta.varberg.se/
2. Turn on "Fastighet, gräns och beteckning".
3. Zoom in until layer is visible.
4. Testclick on any property, and the info window will appear: image
5. Close the info window.
6. Zoom out to a scale where the layer is NOT visible.
7. Click on the map, and the info window will still appear: 
image

The infoclick should no work on an invisible layer. No info window should appear.

To fix this we might be able to control infoclick so that it's only clickable on a layer that is visible.

OlofSvahnVbg:
closes #1348

### DIFF
--- a/new-client/src/models/Click.js
+++ b/new-client/src/models/Click.js
@@ -18,7 +18,7 @@ function query(map, layer, evt) {
   // is not enough, we must also respect the min/max zoom level settings, #836.
   if (
     layer.layersInfo &&
-    layer.getMinZoom() <= currentZoom &&
+    currentZoom > layer.getMinZoom() &&
     currentZoom <= layer.getMaxZoom()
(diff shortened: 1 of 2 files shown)


OlofSvahnVbg:
Apologies for spelling of logic.

PR #1356, opened 2023-06-21
FME-Server plugin: keep the drawn area

### LINKED ISSUE #1321: FME-Server plugin - Keep Area after order.
Today, when making an order using the FME-server plugin, the selected/drawn area is deleted.
Some users want to order diffrent products using the same selected/drawn area. But its really hard (impossible) to draw the exact same area again.

So it would be really good to be able to keep the selected/drawn area.

In our case it could be forced to keep area because its just to draw a new one if you want to. The old one will just be replaced.

Any opinions?

image

jesade-vbg:
Closes #1321


jesade-vbg:
We've tested and havn't been able to cause any weird side effects.

Hallbergs (approved):
Good!

PR #1363, opened 2023-07-10
copy button to copy the coordinates

ChiragSharma0208:
I have added a copy button o copy the corrdinates.

### DIFF
--- a/new-client/public/appConfig.json
+++ b/new-client/public/appConfig.json
@@ -1,7 +1,7 @@
 {
   "appName": "Hajk",
   "version": "3.12.0-rc.2",
-  "mapserviceBase": "",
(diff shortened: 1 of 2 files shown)


jacobwod (changes requested):
Hi! Can you please:
- remove `appConfig.json` from your PR? There's nothing in there that needs to be commited for this feature to work.
- consider using the [`sx` prop](https://mui.com/system/getting-started/the-sx-prop/) on line 240, rather than the `styling` variable?

Thank you.

jacobwod:
One more thing: I'm not sure that floating values are the intended use for `md` prop, as well as the other breakpoint props. They're supposed to specify the number of columns and that number should be an integer. https://mui.com/material-ui/api/grid/#Grid-prop-md

PR #1371, opened 2023-08-15
Fix to correct incorrect sorting when toggling layers in a hajk group…

### LINKED ISSUE #1362: Drawing order Hajk group layer
We have found a bug with the drawing order for hajk group layer. The example of how to reproduce it is from Halmstads ÖP (our isn't external yet) 

To Reproduce
1. Go to https://framtidsplan2050antagande.halmstad.se/ 
2. Zoom in to Halmstad center and turn on the group Trafik under Mark- och vattenanvändning.
3. Look how Vägreservat is below the lines for Ny gång- och cykelväg and Trygg förbindelse som korsar barriär. 
image

4. Turn off Vägreservat and turn it on again. The drawing order has changed. 
image

We would like the drawing order for hajk group layers to be as it is from the start and not change. 

Let me know if you want me to explain this better or if you have any questions!

jesade-vbg:
Works on my computer :) 
Closes #1362

### DIFF
--- a/new-client/src/plugins/LayerSwitcher/components/LayerGroupItem.js
+++ b/new-client/src/plugins/LayerSwitcher/components/LayerGroupItem.js
@@ -101,6 +101,7 @@ class LayerGroupItem extends Component {
     const { layer } = props;
     const layerInfo = props.layer.get("layerInfo");
     this.state = {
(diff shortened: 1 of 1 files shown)


jacobwod (approved):
Looks good 👍

PR #1383, opened 2023-08-25
Feature/1367 permanent drawer fix

### LINKED ISSUE #1367: Fix regression introduced in #1316: drawer buttons disappear making switching between drawer content impossible
I did some more testing  and I'm not sure we can release this in 3.13. The thing is that **if this option is enabled, there is no way to switch between the different content types in Drawer**. Keep in mind that certain maps (like https://framtidsplan2050.halmstad.se) use Drawer to display contents from both the Tools menu _and_ DocumentHandler. (There are also ideas about displaying legend graphic in a third tab in Drawer.)

So, enabling this option removes also these buttons:
Skärmavbild 2023-06-29 kl  14 56 13

Which unfortunately breaks certain map configurations and user are stuck in this view: 
Skärmavbild 2023-06-29 kl  14 58 03

I can see where it would be desirable to use this option (e.g. don't allow hiding the Drawer) but still be able to switch between Drawer content. 

**As I stated before, this feature is currently half-baked and must be completed before the next release. It is already merged to `develop` so I open a new issue (this) which can be closed upon completion. There's still plenty of time to do this before the next release so I think that we can avoid a revert in this case.**

Ping @OlofSvahnVbg, @Hallbergs, @jesade-vbg.

_Originally posted by @jacobwod in https://github.com/hajkmap/Hajk/issues/1316#issuecomment-1613142389_

OlofSvahnVbg:
closes #1367

### DIFF
--- a/new-client/src/components/App.js
+++ b/new-client/src/components/App.js
@@ -700,7 +700,9 @@ class App extends React.PureComponent {
       if (v !== null) {
         this.setState({ drawerVisible: true, activeDrawerContent: v });
       } else {
-        this.globalObserver.publish("core.hideDrawer");
+        if (!this.state.drawerStatic) {
+          this.globalObserver.publish("core.hideDrawer");
+        }
       }
     });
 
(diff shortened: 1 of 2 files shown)


jesade-vbg:
@OlofSvahnVbg: Looks good!
@jacobwod: Would be good if you could verify that it is working as you expect since you had trouble in ÖP if I remeber correctly.

jacobwod (approved):
Works like a charm, thanks!

Issue #1391, opened 2023-08-31
Invalid plugin configuration in backend example configs


sweco-semara:
Hajk backend map config examples contains an invalid plugin configuration. Specifically "bookmark" plugin, which does not exist.

**To Reproduce**

1. Use either .NET or NodeJS backend-based config with `backend/mapservice/App_Data/map_1.json` or `new-client/public/simpleMapConfig.json` as base
2. Open browser console
3. (Re-)load the map
4. Check browser console

**Expected behavior**
Quiet/empty console on using Hajk-supplied config examples. (_Currently also expects an empty warning, see also #1390, but the point in this bug report is that a specific error report is not expected after reproducing per above..._)

**Actual**
Console warning:
> App.js:396 The map configuration contains unavailable plugins: bookmark. Please check your map config and buildConfig.json.  

**Root Cause Analysis**
I think this might be a Hajk 2 leftover in the demo/example config? The bookmarks plugin in Hajk 3 seems to be introduced in #655 and from what I can see it has always been named with a plural "s"?

**Suggested Solution**
Simplest fix is probably to just rename "bookmark" to "boomarks" in both of the following example configs?
1. .NET backend sample config map_1.json [L437](https://github.com/hajkmap/Hajk/blob/master/backend/mapservice/App_Data/map_1.json#L437) in current master
1. NodeJS backend sample config map_1.json [L440](https://github.com/hajkmap/Hajk/blob/master/new-backend/App_Data/map_1.json#L440) in current master

(Not backwards compatible as a new plugin would then start showing in the demos, so another way to get rid of the warning is to just remove the whole plugin/tool config for "bookmark" from both files... But I think Hajk does not need to maintain a status quo for demo configs?)

jacobwod:
Hi @sweco-semara, 

I absolutely agree. In fact, I've fixed it (outside of upstream though, so you had almost no chance finding it) in b10b50afdbebefa4161c6aed995e4413c413b00a and 89cdefa095dc94ba4f5befc66d3e1249bd3dd5c6. 

The first commit ensures that `preset` is forced into the array we check against. `preset` is a special case as it can exist in the map config object but can never be a part of the `activeTools` object and will thus always show up as a warning without this. 

The second commit ensures that we don't print an empty message.

Feel free to implement your solution though, as mine was never part of the main branch. Thanks.

PR #1392, opened 2023-09-04
Fix: Prevent console-warning on correct plugin config

### LINKED ISSUE #1390: Client unsupported plugins check assumes map configuration contains at least one missing plugin
Hajk client will currently always log the silent warning re. missing plugins in map config to the console, regardless if the config is clean/only contains active tools.

This bug is cosmetic only, mostly for developers or admins wanting to check that the Hajk config is correct. Current state is a bit confusing, but the map will still load without issues.

**To Reproduce**
1. Clean up a map configuration, make sure no inactive or invalid plugins are referenced
2. Open the browser console
3. (Re-)load the map
4. Check browser console

**Expected behavior**
Expect console to be quiet/empty if no errors.

**Actual**
> The map configuration contains unavailable plugins: . Please check your map config and buildConfig.json.

**Suggested fix**
It seems there is a hard assumption in App.js that const `unsupportedToolsFoundInMapConfig` will always be set/contains at least one value. Suggested fix is to guard the join + console log operation on [L397](https://github.com/hajkmap/Hajk/blob/master/new-client/src/components/App.js#L397) in current master to not log if unsupportedToolsFoundInMapConfig is indeed empty.

sweco-semara:
git cherry-pick of @jacobwod existing fixes to prevent "false flag" console logging re. invalid plugin config
Fixes #1390
Cherry picks:
* https://github.com/hajkmap/Hajk/commit/b10b50afdbebefa4161c6aed995e4413c413b00a
* https://github.com/hajkmap/Hajk/commit/89cdefa095dc94ba4f5befc66d3e1249bd3dd5c6

### DIFF
--- a/new-client/src/components/App.js
+++ b/new-client/src/components/App.js
@@ -449,11 +449,12 @@ class App extends React.PureComponent {
       });
 
     // Display a silent info message in console
-    console.log(
-      `The map configuration contains unavailable plugins: ${unsupportedToolsFoundInMapConfig.join(
-        ", "
-      )}. Please check your map config and buildConfig.json.  `
-    );
+    unsupportedToolsFoundInMapConfig.length > 0 &&
+      console.info(
+        `The map configuration contains unavailable plugins: ${unsupportedToolsFoundInMapConfig.join(
+          ", "
+        )}. Please check your map config and buildConfig.json.  `
+      );
   };
   /**
    * @summary Initiates the wanted analytics model (if any).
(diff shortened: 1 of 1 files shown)


jacobwod (approved):
Confirmed working, I've had this in production for about two weeks. Does not solve the `bookmark` vs `bookmarks` issue you mention @sweco-semara, but I think you've solved it elsewhere already.

sweco-semara:
Correct @jacobwod, I created a separate issue for `bookmark` vs. `bookmarks` - see #1391 and corresponding PR/fix [#1393](https://github.com/hajkmap/Hajk/pull/1393).

Hallbergs (approved):

PR #1393, opened 2023-09-04
Fix: Correction of invalid plug-in sample in back-end config

### LINKED ISSUE #1391: Invalid plugin configuration in backend example configs

sweco-semara:
Fix invalid demo config syntax, enable the bookmarks plug-in ("Bokmärken").
Resolves #1391

### DIFF
--- a/backend/mapservice/App_Data/map_1.json
+++ b/backend/mapservice/App_Data/map_1.json
@@ -434,7 +434,7 @@
       "index": 15
     },
     {
-      "type": "bookmark",
+      "type": "bookmarks",
       "options": {
         "target": "toolbar",
         "instruction": "",
--- a/new-backend/App_Data/map_1.json
+++ b/new-backend/App_Data/map_1.json
@@ -437,7 +437,7 @@
       "index": 15
     },
     {
-      "type": "bookmark",
+      "type": "bookmarks",
       "options": {
         "target": "toolbar",
         "instruction": "",


jacobwod:
I didn't find any active uses of `bookmark` in the config files I've searched through (quickly), but I wonder (from top of my head, really - haven't had time to run any code today): could this break existing configurations? Doesn't the actual Bookmark plugin need to be re-written, to reflect the name change of property's key and read its options from the new key?

sweco-semara:
@jacobwod I had the same doubts before changing this, but it's non-intrusive for any existing production or test config. The issue is only present in Hajk's example configuration - the distribution used to contain the intention to enable "bookmark" and there is no such plug-in. Effect: Running Hajk's exmple according to the README guide will always render dead config + console warning. The PR will just rename demo config to the correct plug in so that it shows in a demo setting. Does not affect any real config. I have tested it and the bookmarks plug-in will appear in the map tools. (I.e. PR does not change any plug-in name, it's the other way around: it changes the intention / syntax of the provided demo/examples only.)

Hallbergs (approved):

PR #1395, opened 2023-09-06
The new dotnet backend

sweco-semtto:
The new dotnet backend.


jacobwod:
Thanks for the review inquiry. Our organisation has fully moved away from the .NET architecture and I'm not able to do a qualified review on it.

backend-dotnet/MapService/Controllers/MapConfigController.cs
```diff
@@ -0,0 +1,697 @@
...
+            JsonDocument mapDocument;
+
+            try
+            {
+                if (AdHandler.AdIsActive)
+                {
+                    var adHandler = new AdHandler(_memoryCache, _logger);
+
+                    string? remoteIpAddress = adHandler.GetRemoteIpAddress(HttpContext);
+                    if (!adHandler.IpRangeRestrictionIsSet())
```
jesade-vbg, line 113: If I have IP-range set to null in config and X-Control-Header to null I will never be able to get the map. Today, I don't use any proxy or X-Control-Header. Maybe I got something backwards...........

Hallbergs: Just had a quick look at the code and i think you are right. If you use windows-authentication we still check the ip-restriction, which should only be done when grabbing the username from the header, right?

jesade-vbg: Yes, just to clarify:
```
"LookupActive": true,
"IdentifyUserWithWindowsAuthentication": true,
"TrustedProxyIPs": null,
"TrustedHeader": null,
```

Hallbergs: @sweco-semtto Do you have any input on this? Did you test the AD-connection with windows authentication? If so, could you guide @jesade-vbg on how it should be set up?

sweco-semtto: We tested the AD connection using Windows authentication. Correct me if I'm wrong, but I don't see a valid case where TrustedHeader is set to null. I will be more than happy to guide @jesade-vbg on how set this up.

Hallbergs: We're not fetching the user name from a header-key if we're using Windows authentication?

```c#
public string PickUserNameToUse(HttpRequest request, string? userName)
{
    if (IdentifyUserWithWindowsAuthentication)
        return GetWindowsAuthenticationUserName(request.HttpContext);
    else
        return GetValueFromTrustedHeader(request, userName);
}

public string GetWindowsAuthenticationUserName(HttpContext? httpContext)
{
    if (httpContext?.User?.Identity?.IsAuthenticated == true)
    {
        var userName = httpContext.User.Identity.Name;
        var userNameWithoutDomain = userName.Split('\\').Last();

        _logger.LogInformation("Authenticated user: {0}", userNameWithoutDomain);

        return userNameWithoutDomain;
    }
    else
    {
        _logger.LogInformation("User is not authenticated.");
        return string.Empty;
    }
}
```

I don't get why the trusted header should be set when we're dealing with the Windows authentication approach. Am i missing something? I think the code above suggests that the trusted header can be null in the Windows authentication case.

Hallbergs: Also, i don't think the nullish `TrustedHeader` is the main problem  but rather why the ip-restriction must be set even when we're using Windows authentication.

jesade-vbg: @sweco-semtto I can't actually see why I would need a header with user and IP-restriction when IIS is exposed at 443 and handling users with windowsAuth. Either way.... I don't have anything to give me any X-Control-Header.

linusfj:
I have tested the new version of the .NET backend, specifically the endpoints for `admin`. I found issues with certain endpoints, potentially due to my setup.

Notably, I get a 500 error when trying to create, update, or delete a map config in `mapsettings`.

sweco-semtto:
> I have tested the new version of the .NET backend, specifically the endpoints for `admin`. I found issues with certain endpoints, potentially due to the setup.
> 
> Notably, I get a 500 error when trying to create, update, or delete a map config in `mapsettings`.

Thank you for your feedback. Are you sure you have the correct permissions?

sweco-semtto:
Great, well done!

Hallbergs:
For information: We're awaiting merge of #1437 before handling this PR, as decided in the "code group".

jacobwod:
Converted PR to draft to reflect the current status

jacobwod:
Due to recent development and fixes, this branch will now replace the legacy .NET 4.5 in the main `develop` branch. The legacy backend can be found in the `legacy-dotnet-4.5-backend` branch. 

There are still unresolved issues in this backend implementation, notably lack of logging and the requirement for Windows as runtime platform.

Hallbergs (approved):
Approved, see: https://github.com/hajkmap/Hajk/pull/1395#issuecomment-1982874607

PR #1400, opened 2023-09-09
Support relative URLs

### LINKED ISSUE #1399: Support for relative URLs
Make Hajk fully support relative URL:s as discussed in https://github.com/hajkmap/Hajk/discussions/1396.

Identified spots in Hajk develop around v3.12 that need attention for implementing this:
1. Layerswitcher's GetLegendGraphic functionality. ConfigMapper.js, mapWMSConfig inner function [getLegendUrl](https://github.com/hajkmap/Hajk/blob/develop/new-client/src/utils/ConfigMapper.js#L65) - `protocol` handling currently does not support relative URL:s. If using relative links for WMS URL, e.g. layers.json WMS config `"url": "/geoserver/wms"` will result in invalid links being constructed for getting the legend image - `https://geoserver/wms?`
2. PrintMode's printing functionality. PrintModel.js, prepareImageLayer, [setImageLoadFunction](https://github.com/hajkmap/Hajk/blob/develop/new-client/src/plugins/Print/PrintModel.js#L984) function call on the layer source sets a function that cannot execute with a relative URL because of the `new URL(src);` call. This will result in an exception,  only visible in the console - the printing stops without user notice (spinner continues forever).

sweco-semara:
Realizes https://github.com/hajkmap/Hajk/discussions/1396 / resolves #1399.

Tested with:
* layerswitcher
* print
* search
* infoclick
* location
* sketch
* measure
* anchor
* coordinates

### DIFF
--- a/new-backend/server/common/server.js
+++ b/new-backend/server/common/server.js
@@ -292,7 +292,7 @@ built-it compression by setting the ENABLE_GZIP_COMPRESSION option to "true" in 
    * - /api/v1/proxy/geoserver
    * - /api/v2/proxy/geoserver
    * @issue https://github.com/hajkmap/Hajk/issues/824
-   * @issue https://github.com/hajkmap/Hajk/issues/1390
+   * @issue https://github.com/hajkmap/Hajk/issues/1309
    * @returns
    * @memberof ExpressServer
    */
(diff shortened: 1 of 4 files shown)


Hallbergs (approved):
Looks good!

I haven't tested it with relative url:s - but it works as expected with regular url:s.

sweco-semara:
@Hallbergs I have tested with relative URL:s in two environments using different "base" URL:s, so with your absolute URL testing this gives a solid test coverage. The upside of this patch/PR is a fully portable Hajk config where layers+map config JSON can be copy/pasted between stage or acceptance testing and production environments. I suppose it also simplifies in some container scenarios for Docker et.al. Thanks for review and merge!

Issue #881, opened 2021-10-19
Subfolder for the json-documents to "Dokumenthanteraren 2.0"


NiklasErikssonOrebro:
**Is your feature request related to a problem? Please describe.**
In order to be able to organize the json documents used for ”Dokumenthanteraren 2.0” into more ”Berättarkartor” than just one, it would be good if the json-documents used can be structured on the server (… \ App_Data \ documents) under different folders from outside which ”Berättarkarta” they belong to.

**Describe the solution you'd like**
Today there is a start function under the tab for ”Dokumenthanteraren 2.0” called ”Generella dokumnetinställningar”.
Instead, I would like to see that function come up as number 2 and that you instead start by choosing which " Berättarkarta” I want to work with", see picture below.
When a new " Berättarkarta" is created, a new folder is also created on the server under the directory  …\App_Data\documents,  eg  (…\App_Data\documents\berattarkartaXXX), with a link to the " Berättarkarta" created so the json-documents how creates later in step two (the step that is today step 1) is stored on the server under this folder.

**Additional context**
![Bild](https://user-images.githubusercontent.com/92783874/137882129-08bfbd8e-9341-44ff-863a-5536314b1143.PNG)

Anyou22:
[https://github.com/hajkmap/Hajk/blob/feature/orebro_tsm_3/new-backend/server/common/api.yml](url)

Anyou22:
_[https://github.com/hajkmap/Hajk/tree/feature/orebro_tsm_3](url)_ 
We made the following changes in:
-New-backend: we added 2 api:s for creating folders and showing the folder lists. Also, we have changed the previous api:s as load, delete, list, save and create.

New-admin: now it is possible to create a new folder or choose a specific folder. Create and list a document under a specific folder. 
Also, we can connect the map to the folder and document. 
We updated the documenthandler2.0 tool by creating a link which is connected to new-client. An admin can specify a link, document(by choosing a specific folder) or map.

New-client: A client can select a document from a specific folder by clicking on Menu and then a document will appear.

Anyou22:
**Testing the old new-backend with the new version of new-admin and new-client**
Documenthandler2.0 crashed once we run the new-admin but new-client tool works without any errors.

PR #1402, opened 2023-09-09
Feature/orebro tsm 3

Sunix71:
We have made an attempt to develop a solution for issue 881 (https://github.com/hajkmap/Hajk/issues/881). We have added folder functionality to Dokumenthanterare 2.0 so that users can create folders and place their documents in these folders or in the root folder if desired (…\App_Data\documents\).

This has meant that we have had to change the API's endpoints so that it also searches for a directory and not just a document.
We have done this in a way so it should not affect existing configurations that are already set up.

Since we have 2 backends, we have made it such that the administrator needs to set a configuration-key in config.json (new-admin\public\config.json) for this functionality to work.
The key looks as follows: "use_document_folders": true
And should be located under Documenthandler and mappsettings in the config file. The system should work without this key, but then without the functionality.

The map configuration file has received an additional attribute named folder, it looks like this:
{
              "title": "Exempeldokument2",
              "folder": "",
              "document": "exempeldokument2",
              "color": "",
              "icon": {
                "materialUiIconName": "",
                "fontSize": "large"
              },
              "maplink": "",
              "link": "",
              "menu": []
            }
However, the system should function without this attribute.


new-client/src/plugins/DocumentHandler/DocumentHandlerModel.js
```diff
@@ -235,12 +238,19 @@ export default class DocumentHandlerModel {
     }
   }
 
-  async fetchJsonDocument(title) {
+  async fetchJsonDocument(folder, title) {
     let response;
     try {
```
jesade-vbg, line 243: Please exclude hfetch from IF duplicating if. Logic first, one hfetch after. 
```
let url = `${this.mapServiceUrl}/informative/load/`;

if(folder){
  url += `${folder}/${title}` 
}else{
  url += title;
}

hfetch..........
```

jesade-vbg (changes requested):
I've had a quick look and noticed some stuff.
Note that I have not yet tested.

jesade-vbg:
> I've gone through all your comments and hope I haven't missed anything. Made a new push now. Grateful for your time!

Great. And you noticed my url examples wasn't optimized aswell, sorry :) Good.

new-backend/server/apis/v1/services/informative.service.js
```diff
@@ -15,15 +16,29 @@ class InformativeService {
...
-      // Parse the file content so we get an object
-      const json = await JSON.parse(text);
-      return json;
+      if (folder) {
+        file += ".json";
+        // Open file containing our store
+        const pathToFile = path.join(
+          process.cwd(),
+          "App_Data/documents/" + folder,
+          file
```
jacobwod, line 27: I just noticed one more thing that I don't think has been mentioned before. Please use `path.join()` instead of assuming that `/` is the character that we should join folders by. The solution, as written now, is not OS-independent.

jacobwod (changes requested):
Over all, it seems as you've only put this into the `v1` API. I'm not sure if it should be added to V1 at all (as we've recently stated that V1 should maintain compatibility with the old .NET backend), but if you feel like it's good, go for it. Nevertheless, this should also be added to the `v2` part of `new-backend`. 

Don't forget to merge in the latest changes in `develop` before you begin.

new-backend/server/apis/v1/services/informative.service.js
Sunix71: Thank you for your review! We will look into this as soon as possible. We will get back to you about both V2 and path.join().

Sunix71:
> Over all, it seems as you've only put this into the `v1` API. I'm not sure if it should be added to V1 at all (as we've recently stated that V1 should maintain compatibility with the old .NET backend), but if you feel like it's good, go for it. Nevertheless, this should also be added to the `v2` part of `new-backend`.
> 
> Don't forget to merge in the latest changes in `develop` before you begin.

I have added the features to V2 now and pushed (merged develop before I started). I have not yet removed the functionality from V1; that will be the next step, I think. Then, I need to look into path.join to see how one might construct the paths in a different way?! If you have any suggestions please respond :-)

jacobwod:
Sure!

Search and [you will find](https://nodejs.org/api/path.html#pathjoinpaths). 😉 

Basically: don't assume that the delimiter is a `/`, use `path.join()` where you'd use a `/`.

Sunix71:
I've restored V1 and removed / and used path.join instead. There are perhaps more issues to resolve. Thank you for your time!

new-backend/server/apis/v2/services/informative.service.js
```diff
@@ -15,15 +15,36 @@ class InformativeService {
...
+          process.cwd(),
+          "App_Data",
+          "documents",
+          file
+        );
+        const text = await fs.promises.readFile(pathToFile, "utf-8");
+        // Parse the file content so we get an object
+        const json = await JSON.parse(text);
+        return json;
+      }
```
jacobwod, line 47: Hi guys. You seem to duplicate a lot of code, where the only difference is whether you join with the `folder` or not. Can you refactor?

new-backend/server/apis/v2/services/informative.service.js
```diff
@@ -41,17 +62,29 @@ class InformativeService {
...
+          documentName
+        );
+      } else {
+        pathToFile = path.join(
+          process.cwd(),
+          "App_Data",
+          "documents",
+          documentName
+        );
+      }
```
jacobwod, line 87: You seem to duplicate a lot of code here too.  It's okay to use an empty string in `path.join`, so just ensure that `folderName` isn't `undefined` and refactor this part too.

new-backend/server/apis/v2/services/informative.service.js
```diff
@@ -71,6 +104,37 @@ class InformativeService {
...
+  /**
+   * @summary Create a new, empty folder.
+   *
+   * @param {*} folderName to be created
+   * @returns
+   * @memberof InformativeService
+   */
+  async createFolder(folderName) {
+    try {
+      const foldern = {
```
jacobwod, line 116: Please rename this variable.

new-backend/server/apis/v2/services/informative.service.js
```diff
@@ -71,6 +104,37 @@ class InformativeService {
...
+        foldername: folderName, // Return
+      };
+      // …and create a new path to that folder.
+      const pathToFolder = path.join(
+        process.cwd(),
+        "App_Data",
+        "documents",
+        folderName
+      );
+      console.log(pathToFolder);
```
jacobwod, line 126: Remove

jacobwod (changes requested):
I left some comments in code, but it's, mostly the same problem and you have it in more places than those I've marked.

We've tested the branch though and it works fine, so as soon as the code issues are fixed, this should be ready for merge.

Sunix71:
I have gone through the code and made some changes, please have a look through it again, thank you!

Hallbergs (approved):

jacobwod (approved):

jacobwod:
While reviewing and testing, we noticed that it seems impossible to set starting document if that document is in a folder. Perhaps there's another way to do than what we've tried, if so please document it (in Admin). Thanks @Sunix71.

image

PR #1408, opened 2023-10-03
Feature/1404 subchapter level setting bug

### LINKED ISSUE #1404: Subchapter Level Setting Bug
The Subchapter Level setting doesn't work correctly. If you choose, for instance, 0 in the input... 
image
the value does not carry over to client side.

The problem seems to be here (in the code):
image
The property tableOfContent does not appear on the modeldoc object (document).

It does, however, exist in `this.props.options?.tableOfContents`. And if you use this instead, the setting works from admin.

Is this an okay solution, or does it interfere with the node backend perhaps?

OlofSvahnVbg:
closes #1404


jesade-vbg (approved):
Thumbs up

PR #1410, opened 2023-10-04
Features/1409 matomo in node backend

### LINKED ISSUE #1409: Add matomo support to node backend
Looks like most of it are already there (commented), just need to check, verify and maybe do minor changes.

jesade-vbg:
The commented theoretical code worked just fine @jacobwod :) So I just uncommented and added example config to .env.
Closes #1409


Hallbergs (approved):
LGTM

PR #1417, opened 2023-10-11
Feature/1416 fir endpoints in node backend

### LINKED ISSUE #1416: node backend: Add FIR plugin endpoints
I'm currently testing node backend and I require the missing FIR plugin endpoints. They are currently only availiable in the old .Net-backend.

Note that the functionality is ported (no rethinking) from the old.Net backend so these endpoints will be drop-in replacements and will only requires 2 settings in .env and 2 url changes in map config. No changes in client.

/fir/realestateownerreport
/fir/residentreport

Future rethinking will most certainly happen.

jesade-vbg:
Closes #1416

### DIFF
--- a/new-backend/.env
+++ b/new-backend/.env
@@ -213,3 +213,6 @@ OPENAPI_ENABLE_RESPONSE_VALIDATION=false
 #ANALYTICS_TYPE="matomo" 
 #ANALYTICS_OPTION_TRACKERURL=https://xxxxxxxxxx.matomo.cloud
 #ANALYTICS_OPTION_SITEID=4
+
+# FIR_REALESTATEOWNERREPORT_URL=https://fastighetsrapport-utv.domain.se/Report/fastighetsforteckning/xlsx
+# FIR_TEMP_OUTPUT_DIR=/temp/
--- a/new-backend/.env
+++ b/new-backend/.env
@@ -214,5 +214,8 @@ OPENAPI_ENABLE_RESPONSE_VALIDATION=false
 #ANALYTICS_OPTION_TRACKERURL=https://xxxxxxxxxx.matomo.cloud
 #ANALYTICS_OPTION_SITEID=4
 
+# *** FIR SETTINGS ***
+# The url is an endpoint in myCarta FR Direkt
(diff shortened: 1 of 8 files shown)


new-backend/.env
```diff
@@ -213,3 +213,6 @@ OPENAPI_ENABLE_RESPONSE_VALIDATION=false
 #ANALYTICS_TYPE="matomo" 
 #ANALYTICS_OPTION_TRACKERURL=https://xxxxxxxxxx.matomo.cloud
 #ANALYTICS_OPTION_SITEID=4
+
+# FIR_REALESTATEOWNERREPORT_URL=https://fastighetsrapport-utv.domain.se/Report/fastighetsforteckning/xlsx
+# FIR_TEMP_OUTPUT_DIR=/temp/
```
jacobwod, line 218: Sorry to by picky but can you add some comments about the recent additions in .env? Take a look at the sections above and do something similar. :)

new-backend/server/apis/v2/router.js
```diff
@@ -5,11 +5,13 @@ import mapconfigRouter from "./controllers/mapconfig/router.js";
 import settingsRouter from "./controllers/settings/router.js";
 import informativeRouter from "./controllers/informative/router.js";
 import adRouter from "./controllers/ad/router.js";
+import fir from "./controllers/fir/router.js";
```
jacobwod, line 8: Please rename to `firRouter`

jesade-vbg:
I've refactored allot. Service added (as it should have been from the beginning). Fixed the other stuff in th PR comments.

jesade-vbg:
Pushed fixed issues and refactored everything. It now uses a service as it should have done from the beginning.

jacobwod:
OK I'm looking into it!

jacobwod (approved):
I'd say go for it. There is a problem with the approach used, but this PR is not the root cause of the problem, namely that we should find a way to conditionally enabled endpoints (as well as do dynamic imports), only when needed. E.g. this `/fir`, or the `/ad` endpoint should not be created by the backend if neccesary keys aren't set in `.env`. But it's outside the scope of this PR.

PR #1418, opened 2023-10-13
Feature/1414 drawer static button bug

### LINKED ISSUE #1414: Document Handler - Drawer Static Button Bug
If you choose to have the drawer permanently locked in place...
image

... an inactive button still appears:
MicrosoftTeams-image

There is logic implemented to keep this from happening, and the problem does not seem to appear in kommungis-utv...
image

... and I'm not sure why this problem occurs, actually. Perhaps something wrong with `drawerButtons` in `App.js`.

OlofSvahnVbg:
closes #1414

### DIFF
--- a/new-client/src/components/App.js
+++ b/new-client/src/components/App.js
@@ -1038,11 +1038,19 @@ class App extends React.PureComponent {
     );
   };
 
-  checkDrawerButtons() {
-    return (
-      !this.state.drawerStatic ||
-      (this.state.drawerStatic && this.state.drawerButtons.length > 1)
+  showDrawerButtons() {
+    // We check if the plugin button (or any button) is empty and then subsequently hidden
+    const hasHiddenPlugin = this.state.drawerButtons.some(
+      (db) => db.hideOnMdScreensAndAbove
     );
+
+    // We want to check if there's only one visible drawerButton
+    const hasSingleVisibleButton =
+      this.state.drawerButtons.length === 1 ||
+      (this.state.drawerButtons.length === 2 && hasHiddenPlugin);
+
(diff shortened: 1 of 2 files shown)


jesade-vbg (approved):

new-client/src/components/App.js
```diff
@@ -1038,11 +1038,23 @@ class App extends React.PureComponent {
...
+
+    // We check if the plugin button (or any button) is empty and then subsequently hidden
+    const isHiddenPluginPresent = drawerButtons.some(
+      (button) => button.hideOnMdScreensAndAbove
     );
+
+    // We want to check if there's only one visible drawerButton
+    const isOnlyOneButtonVisible =
+      drawerButtons.length === 1 ||
+      (drawerButtons.length === 2 && isHiddenPluginPresent);
```
jacobwod, line 1054: Hi! This slipped my attention prior the merge, but anyway: how about a situation which more than 2 drawer buttons? I know that we currently only have two of them, but since any plugin can add such a button (check out the Dummy plugin…) it's not impossible to end up with 3 or more buttons in the future. Have you checked for compatibility in such a situation? Thanks! /J

OlofSvahnVbg: Hello! I think in a situation where we have 2 or more visible buttons (and if you have 3 buttons they probably are all visible) the buttons should not be removed if the drawer is static. This logic only ensures that if there is one button, it doesn't show, OR if there are two buttons (one visible and one invisible) they dont show.

OlofSvahnVbg: But might be an idea to change logic anyways in the event of something like that :)

Issue #1380, opened 2023-08-24
Add Quick Layers to LayerSwitcher


linusfj:
### Overview
The implementation of `Quick Layers` will include enhancements within the `admin` interface, user interface, and the `LayerSwitcher` tab.

Related issues: #1237, #1257, #1284, #1296, #1347 and #1365.

---

### Quick Layers
`Quick Layers` represent a specific grouping within the `LayerSwitcher`, offering quick access to often-used layers. They can be manually added or loaded from pre-defined packages. They provide the advantage of including collections of layers without having to display all of them from the start. `Quick Layers` also offer options to save, edit, load, and manage layers more efficiently.

### Search functionality
The addition of a search function to the `LayerSwitcher` will enable users to locate specific layers or groups. The design will include matching search text with layer and group names, filtering through all tree levels, and the inclusion of a "Clear" button to reset the layer list.

### Snackbar notification
Changes to the `Snackbar` notification system will group notification messages together when toggling either individual layers or groups.

### Layer visibility
Enhancements to layer visibility will include the utilization of bold font to highlight the titles of active layers within the `LayerSwitcher`.

### Draw Order tab
The `Draw Order` tab functionality will allow users to control and visualize the drawing order of layers and background layers.

### Admin UI
Upgrades to the `admin` interface will include the introduction of a section specifically for managing `Quick Layers`. The features for this section includes the ability to input, validate, and manage `Quick Layers`, and functionality for adding and deleting them.

PR #1419, opened 2023-10-19
Feature/1257 highlight active layers

LeeBadal:
# Feature/1257

Improve visibility between selected/not-selected in LayerSwitcher and active/non-active layers in bottom Snackbar.
## Screenshots
Adds a bold font style to the text of selected active layers in the layer switcher:
image

Adds a bold font to snackbar for selected layers, toggle is triggered by clicking the visibility icon.
image

## How:
**LayerGroup/LayerItem** Implements a function getCaption(caption,visible) that returns the appropriate styled component depending on visible argument.
**BreadCrumb.js** (bottom snackbar) uses the **hidden** variable which is related to the visibilityIcon toggle.

Would have preferred to use the same pattern/reuse function, however implemented in the logic similar to component-related code to preserve file-level consistency.


Hallbergs:
@linusfj can you take a look at this? I saw that you opened the issue a while back. Could this be incorporated into the new layerSwitcher that your organization is working on?

linusfj:
We've developed a similar solution on our branch for the new `LayerSwitcher` (#1380). In our implementation, we've emphasized layer and group names in the `LayerSwitcher` using bold text. Your suggestion to extend this bold styling to breadcrumbs as well is a good addition.

I suggest we approve this pull request and subsequently integrate these changes into the new `LayerSwitcher` branch.

linusfj (approved):

Hallbergs (changes requested):
Wouldn't it be easier to put the visible check directly in the style-tag? That way we could avoid alot of code. Something like:
```js
<HeadingTypography sx={toggled ? { fontWeight: "bold" } : null}>
     {this.state.name}
</HeadingTypography>
```
Since you're using a component that already styled (`HeadingTypography`) another clean route could be to put the fontWeight-handling there and pass a prop to control it.

jacobwod:
I agree with @Hallbergs. 

Additionally: can you consider making the bold styling "bubble" upwards in the tree, so that a group that has enabled layers also becomes bold? (In your screenshot @LeeBadal, it'd be bolding the _Länsstyrelsen_ group.) 

Thanks!

LeeBadal:
@Hallbergs Agree, done.

@jacobwod 
Sure, 
image
image
image

Hallbergs (approved):

Hallbergs:
Nice, gj!

jacobwod:
Hi @LeeBadal! Does this really work for top-level groups? Cause I'm having with my top-level groups. 

One difference between my top-level groups and other groups further down is the check-all toggle.

Perhaps this requires the toggle to be work?

Skärmavbild 2023-10-30 kl  10 24 43

jacobwod:
Update, I've fixed it - ensure to grab the latest develop where it's been cherry-picked as of 3e2f3391445711e229842b3e4c019b129dedd641.

PR #1434, opened 2023-11-15
Now checks if Export is allowed

maan002:
Fixes 1433

### DIFF
--- a/new-client/src/plugins/Fir/FirExportResidentListView.js
+++ b/new-client/src/plugins/Fir/FirExportResidentListView.js
@@ -372,9 +372,14 @@ class FirExportResidentListView extends React.PureComponent {
     return (
       <>
         <Accordion
-          disabled={this.props.results.length === 0}
+          disabled={
+            this.props.results.length === 0 ||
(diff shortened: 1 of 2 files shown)


jesade-vbg:
Have you tested this?  I dont think you can filter out the export functionality using "residentList".
I think its only possible to Limit per plugin KIR and FIR.

maan002:
Yes, I have tested this and it works perfectly.
The .NET Backend removes the residentList if you do not have that AD group.
See file ConfigController.cs line 399:

```
           var residentList = firTool.SelectToken("$.options.residentList");
            if (residentList != null)
            {
                var visibleForGroups = residentList.SelectToken("$.visibleForGroups");
                if (HasValidVisibleForGroups(visibleForGroups) && !IsGroupAllowedAccess(userGroups, visibleForGroups))
                {
                    (firTool.SelectToken("$.options") as JObject).Remove("residentList");
                }
            }
```

Hallbergs:
Seems like the fact that we have two backends has lead to confusion again! 😖

PR #1436, opened 2023-11-20
Fix layout issue with hidden DrawerToggleButtons

### LINKED ISSUE #1364: Search field location
**Describe the bug**
In maps where you don't have a panel, the location of the search field has previously automatically ended up on the left. As of 3.10, we have noticed that the search field instead ends up on the right in cases where there is a widget enabled. See screenshots below from 3.9 and 3.10 with and without widget.

**To Reproduce**

1. Configure a map with search tools but no panel. Configure another tool placed in a widget.
2. Load the map and check where the search field is placed.
3. Configure a map with search tools but no panel. Configure another tool placed as a control button.
4. Load the map and check where the search field is placed.

**Expected behavior**
The search field should have the same location in the map regardless of whether a widget is activated or not.

**Screenshots**
3.9 without widget – the search bar ends up on the left
image

3.9 with widget – search bar still ends up on the left
image
 
3.10 without widget – the search bar ends up on the left
image

3.10 with widget - the search bar now ends up on the right
image

**Desktop (please complete the following information):**

- OS: Windows 10 Enterprise
- Browser: Microsoft Edge
- Version: 114.0.1823.67

**Additional context**
Would it be possible to add the ability to control the placement of the search field in admin?

linusfj:
Closes #1364.

### DIFF
--- a/new-client/src/components/Drawer/DrawerToggleButtons.js
+++ b/new-client/src/components/Drawer/DrawerToggleButtons.js
@@ -116,8 +116,14 @@ function DrawerToggleButtons({
     );
   };
 
+  // Check if all buttons are hidden on md screens and above.
+  const areAllButtonsHiddenOnMdAndAbove = drawerButtons.every(
+    (b) => b.hideOnMdScreensAndAbove
+  );
+
   return (
-    drawerButtons.length > 0 && (
+    drawerButtons.length > 0 &&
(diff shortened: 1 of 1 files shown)


Hallbergs (approved):
Seems to work as intended! :) @jacobwod @jesade-vbg do you guys want to test before merging?

PR #1440, opened 2023-12-04
Feature/1439 nodebackend consume user groups headers

### LINKED ISSUE #1439: node backend: Consume both user and groups from http headers.
We need a way, a diffrent way, to verify access and layervisibility using headers.

The suggested middle way is to have a proxy that sends both user and groups to nodejs backend using http headers.
Then... the nodejs backend consumes both user and groups from the http headers.

Example input to nodejs. Headers configured in .env-file
```
X-Control-Header:USER
X-Control-Group-Header:Group1,Group2,Group3
```

This way, nodejs does not need the AD connection and search capabilities because this is moved to the proxy.

By implementing this, it might be a sollution to get up and running faster when using IIS webserver.

Not that:
This sollution will also require a working proxy and @Hallbergs has a potential .net core IIS module for this.
The potential proxy will be linked in this issue in the near future.

jesade-vbg:
Closes #1439


jacobwod (approved):
Well done! 👍

PR #1444, opened 2023-12-13
Feature/1443 new filters

### LINKED ISSUE #1443: Infoklick - New Filters
1. One filter in infoklick that returns a specified number if the value is NaN or undef

Example:
{NaN|naNToNum('-1000')}
  outputs: -1000

2. One filter that check if the value is greater than a specified number, see also filter 'lt'

Example:
  {10.3|gt('9.2', 'GreaterThan', 'LessThan')}
  outputs: 'GreaterThan'

3. Enhance the filter 'lt' to check if the 'lessThan' or 'greaterThan' argument is undef (in the case of an empty string)

maan002:
Closes #1443

### DIFF
--- a/new-client/src/components/FeatureInfo/FeaturePropsFilters.js
+++ b/new-client/src/components/FeatureInfo/FeaturePropsFilters.js
@@ -216,6 +216,23 @@ filters.add("gt", function (value, test, greaterValue, lessValue) {
 });
 
 /*
+  naNToNum
+  If value is NaN or undef returns num as a number, otherwise returns value
+  Example:
+  {NaN|naNToNum('-1000')}
+  outputs: -1000
+  {10.3|naNToNum('-1000')}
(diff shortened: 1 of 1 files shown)


jesade-vbg:
Have you tried:
(value in this case is the attribute name from db (3.12))
```
value='10.3'
{value|lt('11', value, 'greaterThan')}
```

And

`{NaN|default('-1000')}`

maan002:
Default doesn't work for the case:

{'Sometimes the value is a string that needs to be converted to number'|default('-1000')}
Outputs: 'Sometimes the value is a string that needs to be converted to number'

nAnToNum works:
{'Sometimes the value is a string that needs to be converted to number'|nAnToNum('-1000')}
Outputs: -1000 (as a number)

jesade-vbg (approved):

PR #1449, opened 2023-12-20
Feature/1438 visible for groups within tool options

### LINKED ISSUE #1438: node backend: visibleForGroups within tool options
A plugin/tool can have nested options that could have visibleForGroups.
These should be able to be restricted/removed by node backend just like a plugin/tool.

The old .Net backend had this functionality on some specific plugins.

jesade-vbg:
Closes #1438

### DIFF
--- a/new-backend/.env
+++ b/new-backend/.env
@@ -36,7 +36,7 @@ PORT=3002
 # The recommended setting is to leave this empty (i.e. outcommented). This means that all
 # currently supported versions of the API are enabled.
 # If you want to enable only specific version(s), supply a comma-separated list, e.g.
-#API_VERSIONS=2,3 # Disables /api/v1, enables /api/v2 and /api/v3
+API_VERSIONS=2 # Disables /api/v1, enables /api/v2 and /api/v3
 
 # When an Express app is running behind a proxy, app.set('trust proxy') should be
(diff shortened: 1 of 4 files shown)


new-backend/.env
```diff
@@ -195,7 +195,10 @@ AD_RECONNECT=true
 
 # For development and testing purposes, a user name can be supplied here.
 # Make sure to always disable this in production!
-#AD_OVERRIDE_USER_WITH_VALUE=someUser
+# AD_OVERRIDE_USER_WITH_VALUE=someUser
+# For development and testing purposes, groups can be supplied here.
+# Make sure to always disable this in production!
+# AD_OVERRIDE_USER_GROUPS_WITH_VALUE=someGroup1,someGroup2
```
jacobwod, line 201: I had a quick look at the comment related to these keys and this got me thinking: should we be a bit harder on this and actually not allow these keys at all (or ignore their values in code), if `NODE_ENV=production` (see line 24). What's your opinion?

jesade-vbg: Yes, its a really good idea actually. Prevents .env accidents even more.

jesade-vbg: Check added:
https://github.com/hajkmap/Hajk/pull/1449/commits/9fd17fa667fc1381b488155b978b27092819ecce

jacobwod (approved):

PR #1456, opened 2024-01-24
Feature/1455 sketch rotate features

### LINKED ISSUE #1455: Sketch plugin: Rotate drawn objects
It would be great if it was possible to rotate drawn objects.

For example: You draw a house using Rectangle and then rotate it correctly in place (Perhaps checking for a good place for Attefall or Friggebod?).

Today its pretty hard to do correctly and you have to use the polygon tool.

Note that I have not thought much about how and where in GUI.

jesade-vbg:
closes #1455

### DIFF
--- a/new-client/src/models/DrawModel.js
+++ b/new-client/src/models/DrawModel.js
@@ -2105,7 +2105,9 @@ class DrawModel {
     });
   };
 
+  // Rotate the currently selected features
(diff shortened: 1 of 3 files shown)


Hallbergs (approved):
Works great. Some style-tags could be removed but don't bother.

Hallbergs (approved):
Perfect!

Issue #1445, opened 2023-12-13
Infoklick - Enhance filter roundToDecimals


maan002:
**Is your feature request related to a problem? Please describe.**
Sometimes a value will be either a string or a number. If the value is a string, you should be able to specify that this is correct and the value should be returned without an error message.

**Describe the solution you'd like**
Add a parameter to roundToDecimals to be able to specify this. The parameter should have a default value that retains the filter's previous functionality.

Example previous functionality:
  {'45.32465456'|roundToDecimals(2)}
  outputs: 45,32

Example new functionality:
  {'Not a number'|roundToDecimals(2,1)}
  outputs: 'Not a number'

PR #1459, opened 2024-01-29
Enhanced filter roundToDecimals to accept a default parameter to prev…

maan002:
…ent error being thrown in case of a string value


maan002:
Closes #1445

Hallbergs (approved):
Make sure to update the documentation!

Issue #1310, opened 2023-03-30
Sketch/Buffer: Make it possible to select Buffer results as polygons in Sketch 


jacobwod:
A feature request came up: users would like to use the Buffer tool and calculate its area. The obvious way would be to use the Measure tool, but it does not support snapping to Buffer objects. Neither does the Sketch plugin.

(NB: The new Measurer tool actually snaps as required.) 

It'd be nice to add this functionality, at least to the Sketch plugin too.
![MicrosoftTeams-image](https://user-images.githubusercontent.com/110222/228774164-22b031f5-1d20-4f16-aa3f-da6ccc2242c2.png)

PR #1460, opened 2024-01-30
Added possibility to buffer from the Sketch plugin

Albinahmetaj:
Issue #1310 

This branch represents the possibility for the user to buffer directly from the sketch plugin as an MUI accordion component.
It follows the same structure as the Buffer plugin with a Buffer component acting as parent, a Buffermodel and a Bufferview. But this time it was developed using functional components.


Hallbergs (dismissed):
Looks good, works great - good job!

Hallbergs (changes requested):
Sorry for keeping you waiting... Haven't really had any time to Hajk the last weeks.

Anyways:
- When enabling the buffer-tool, and clicking a feature in the map, the info-click window opens. This can be avoidied by enabling clicklock when enabling the buffer selection.
- Would it be possible to combine some of the state that are passed via props into a state-object? If you want, we could have a look at it together?

Albinahmetaj:
No problem, I appreciate the review and the comment.

Regarding the changes, sounds great I'll take a look at it.
Absolutely, if you got the time. Just give me a heads up

Hallbergs (changes requested):
See: https://github.com/hajkmap/Hajk/pull/1480#pullrequestreview-1905505250

It would be nice if we could avoid the eslint-disablings. Would it be feasible to destructure here as well?

Hallbergs (approved):

jacobwod (dismissed):

Hallbergs (approved):

PR #1464, opened 2024-02-01
Fix/1463 tooltip captures pointer events

### LINKED ISSUE #1463: Tooltips should not capture pointer events.
May be a bug in MUI or something.
Anyways, I find it very annoying when mouseevents are captured by the tooltip.

Example video attached.

jesade-vbg:
This forced little theme override seems to do the trick. 
Anyone against it?

closes #1463

### DIFF
--- a/new-client/src/components/HajkThemeProvider.js
+++ b/new-client/src/components/HajkThemeProvider.js
@@ -130,8 +130,6 @@ const HajkThemeProvider = ({ activeTools, config, customTheme }) => {
   // to the original one (which requires a re-render).
   const [themeUID, setThemeUID] = useState(Math.random());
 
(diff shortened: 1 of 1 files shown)


Hallbergs:
I think an easier solution would be to set the `disableInteractive` prop on the `<Tooltip />` component? That way we can avoid overrides.

jesade-vbg:
@Hallbergs I looked everywhere for a setting like that :) Why rename it to interactive, when its pointerEvents thats used behind the curtain.......

Anyways. As I've now read that disableInteraction=false is a default setting on purpose to pass WCAG, I will close the Issue and PR and add this to customTheme.json when necessary instead. We can't force changing this default in the Hajk-project.

PR #1474, opened 2024-02-12
InfoClick highlight in collection view.

### LINKED ISSUE #1472: InfoClick: Highlight in collection view.
In the InfoClick listview, features on the map gets highlighted when hovered in the list.

Request has come in to add this highlighting to the collection list view as well.

Due to the fact that it theoretically can be many features to highlight, this function should be optional and throttled if activated.

Example of the feature collection list view for reference.
In this case, the performance problems would be tiny tiny and not noticable.

image

jesade-vbg:
Closes #1472 

I'm aware it's a bit bulky with the throtteling, but I think its worth it.
The feature is disabled by default in admin.

### DIFF
--- a/new-admin/src/views/tools/infoclick.jsx
+++ b/new-admin/src/views/tools/infoclick.jsx
@@ -26,6 +26,7 @@ var defaultState = {
   allowDangerousHtml: true,
   useNewInfoclick: false,
   useNewPlaceholderMatching: false,
+  useLevel1FeatureHighlight: false,
   transformLinkUri: true,
 };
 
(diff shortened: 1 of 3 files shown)


Hallbergs (approved):
Looks good, works great - good job!

jacobwod:
Great, well done @jesade-vbg! Not a request for change, but I wonder about the centering: does it work good by picking center from the first feature's geometry? If not, perhaps you could fit to the extent of the highlight source (I think there's something like `getExtent()` on source, and that can be used in `View.animate()` or `View.fit()`.)

jesade-vbg:
I'll look into it. Maybe use center of layers extent. But I don't think fit() will be good as it also zooms. I'll make some tests, good point.

Issue #1177, opened 2022-09-16
Sketch: allow disabling stroke for polygons and circles


jacobwod:
Sometimes users might want to draw a feature without any border. We could achieve this functionality by either a) allowing a stroke width of 0px, or b) adding a "None" type to the selection (where Solid, Dashed etc are available today).

Hallbergs:
"None" in the stroke-type would be a nice addition.

PR #1480, opened 2024-02-19
Opportunity to disable stroke for drawtypes "Circle" & "Polygon"

Albinahmetaj:
Issue: #1177 

Note: Additionally, I've conducted tests to verify the consistency of the stroke value 'none'/stroke style by exporting to a KML file and importing it across various software platforms, aside from Hajk. This was done to ensure alignment with the expected behavior across different environments.

### DIFF
--- a/new-client/src/plugins/Sketch/components/featureStyle/FeatureStyleSelector.js
+++ b/new-client/src/plugins/Sketch/components/featureStyle/FeatureStyleSelector.js
@@ -19,7 +19,7 @@ export default function FeatureStyleSelector(props) {
       !["Circle", "Polygon"]?.includes(activeDrawType) &&
       drawStyle.strokeType === "none"
     ) {
-      // Update drawStyle with solid strokeType
+      // We want to update the drawStyle with the default values
       setDrawStyle((prevDrawStyle) => ({
         ...prevDrawStyle,
         strokeType: DEFAULT_DRAW_STYLE_SETTINGS.strokeType,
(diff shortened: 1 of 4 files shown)


new-client/src/plugins/Sketch/components/featureStyle/FeatureStyleSelector.js
```diff
@@ -1,12 +1,36 @@
...
+        strokeType: DEFAULT_DRAW_STYLE_SETTINGS.strokeType,
+        lineDash: STROKE_DASHES.get(DEFAULT_DRAW_STYLE_SETTINGS.strokeType),
+        strokeColor: {
+          ...props.drawStyle.strokeColor,
+          a: DEFAULT_DRAW_STYLE_SETTINGS.strokeColor.a,
+        },
+      }));
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [props.activeDrawType, props.drawStyle]);
```
Hallbergs, line 32: I would suggest that you destructure the props outside the useEffect and remove the eslint-disable row!

Something like:

```js
const {activeDrawtype, drawStyle} = props;

useEffect(() => {...}, [activeDrawtype, drawStyle])

Hallbergs (changes requested):
Seems to work great! Just fix the dependency array and it's ready for merge!

new-client/src/plugins/Sketch/components/featureStyle/FeatureStyleSelector.js
Albinahmetaj: Cheers! I'll make sure to remove the eslint-disable rows in #1310 also.

Hallbergs (approved):

PR #1488, opened 2024-03-14
New repo structure

jacobwod:
This project has for a long time in fact been some kind of a monorepo, with multiple applications scattered around the root directory, accompanied by some helper scripts, build files and documentation. I propose we clean this up a bit. The goals are:
- clear structure that mimics other popular projects
- "apps" are grouped together in a directory (please not that "apps" does not necessarily mean "something with an UI" in this context - often repos put their APIs inside the `apps` folder - that what I propose to stick to)
- scripts moved to a separate location
- perhaps some restructuring regarding our Docker build files
- and finally, let's drop the `new-` prefix on our apps

This PR is not ready for a merger yet, please leave your input below or push changes directly to PR branch.

I wouldn't want to complicate any ongoing development or break custom build scripts, so let's collaborate on this one. 🤝 

UPDATE: 

I realise this PR is rather large, so here's a more visual view of the changes as proposed at time of writing.
Skärmavbild 2024-03-14 kl  11 01 41


jacobwod:
I agree @sweco-semara. Regarding the new "home" for our community extensions/plugins - I think it's been overlooked. It's good you mention it. 

I think there are at least two mutually exclusive solutions: either a) in other forks, outside of the main repo (kind of disorganised, people can put their contributions wherever they want), or b) as a top-level `community/` directory (as you propose; I like the _one_ directory, top-level solution too). 

If we go for B: should we allow back-merging of community extensions to the main repo, or just leave `community/` as a placeholder, to make it easier for those forking and building their own custom releases? (I understand it's a big decision, but we can talk about it now and change in the future, if it doesn't work out for some reason. 😄)

---

Just a quick tip for those of you wanting to test this branch: I highly recommend cloning it into a separate directory on your local computer in order to avoid directories being created and disappearing when you switch branches. Next you could run `script/reinstall_modules.sh` as it's already been prepared for this new structure.

Hallbergs (approved):

PR #1489, opened 2024-03-18
Feature/1482 qr code

### LINKED ISSUE #1482: QrCode implementation in Print/Anchor
So as we all now, qr codes are cool, and therefore we should have them in hajk. Specifically in Print and Anchor (Utskrift and Dela).

I've already tested an implementation using the `enableAppStateInHash` and it seems to work fine. 

Qr codes can be implemented in both print:

image
image

And in Anchor:

image

And we should probably only allow qr code generation if  `enableAppStateInHash` is enabled. Question is, what do we show if  `enableAppStateInHash` is false? Nothing, or maybe inactive buttons? What do you think?

OlofSvahnVbg:
closes #1482


jesade-vbg (approved):
Looks good!

PR #1491, opened 2024-03-20
Feature/1220 configurable drawer button

### LINKED ISSUE #1220: Making icon and titles configurable for Kartverktyg
We want to be able to change the icon and text for Kartverktyg.

We suggest making the drawerButtonIcon, drawerButtonTitle and drawerTitle configurable in admin.
The default icon is suggested to be Map and the titles KARTVERKTYG.

linusfj:
Closes #1220.


Hallbergs:
Is it by design you only allow for two icons? If we're making the button configurable, maybe we should allow for any icon?

linusfj:
In the current version, we only allow for two icons, but expanding this to any icon is a good idea for later iterations.

Hallbergs (approved):
OK. I would be much happier to see possibility to choose icon more freely.

Issue #1469, opened 2024-02-08
GeoWebCache: Hajk issues with stand-alone GWC WMS integration


sweco-semara:
**Describe the bug**
Hajk is not able to integrate with stand-alone instances of GeoWebCache, there are a number of assumptions being made that only a GeoServer-bundled GeoWebCache will fulfill. E.g:
* Admin assumes GeoServer REST endpoint availability for listing WMS server workspaces. Workspace list endpoint is not available in GeoWebCache.
* Admin failes to parse GeoWebCache GetCapabilities in full WMS mode to list published projections, resulting in Hajk not able to integrate with non-EPSG:3009 GeoWebCache WMS since projection selection dropdown will be empty in admin and SWEREF 99 TM/EPSG:3009 seems to be assumed regardless of map default projection

**To Reproduce**
1. Start a GeoWebCache stand-alone instance, with full WMS mode and at least one layer with a specific projection
2. Use Hajk admin and try to add the WMS

**Expected behavior**
Hajk admin should reflect GeoWebCache behaviour, i.e:
1. Not enable the user to click "List workspaces" since GWC lacks this capability
1. Be able to parse projections published from capabilities and fill the drop-down for projection selection

**Additional context**
* This issue is opened as a "placeholder issue" and will be detailed as further investigations take place, with the ambition to present a patch to solve these issues.
* _The current idea is to "qualify" GeoWebCache as a distinct server-type in Hajk, this would allow admin users to select "geowebcache" or "geowebcache-standalone" in the Sever Type drop-down, after which admin could adjust further GUI to reflect actual GWC capabilities - this will be discussed with Hajk core developers on upcoming code group meetings_

sweco-semara:
[PR](https://github.com/hajkmap/Hajk/pull/1493) is now ready for review. It includes several improvements for avoiding Hajk admin crashing on all sorts of configuration errors thrown at it during testing, including CORS block / WMS servers missing CORS headers for Hajk which has been a pretty common error in our customer's test and productions environments (before pointing this out).

PR #1493, opened 2024-03-26
Stand-alone GeoWebCache support

sweco-semara:
* Resolves hajkmap/Hajk#1469
* Add support for new WMS server type "geowebcache-standalone"
* Add optional support for configuring a default server type, as well as existing option for default URL (backwards compatible, falls back to "geoserver" if not present)
* Re-name admin WMS option from "GeoWebCache" to tiled (since WMS-C tiled mode has nothing to do with GeoWebCache and it's now a supported server type, making the old admin title for tiled option very confusing)
* Use constants for WMS versions and server types
* Avoid non-conformant WMS v1.3.0 GetCapabilities for geowebcache-standalone server type
* Automatically default to tiled loading for geowebcache-standalone server type
* Spellcheck admin sv-SE labels for queryable layers ("klick")

More robust WMS admin, guard against GetCapabilties load failures or format errors
* Console warning if unsupported spatial reference system found in WMS capabilities
* Fix: Async load could crash WMS admin if clicking too fast on layers/capabilities, not ready for projections update
* Fix: Do not reference undefined objects in admin after failed GetCapabilities
* Fix: Handle failed net requests e.g. due to CORS-block, instead of crashing admin
* Fix: Handle API failures for getting layers config gracefully (e.g. API endpoint version mismatch), instead of crashing admin
* Fix: Handle HTML responses with 200/OK for failed GetCapabilities (like e.g. Hajk development mode and GeoServer proxy)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Core, Admin: Added support for stand-alone GeoWebCache WMS Server. Issue: [#1469](https://github.com/hajkmap/Hajk/issues/1469)
 - Print/Anchor: It's now possible to generate qr codes in Share and Print [#1482](https://github.com/hajkmap/Hajk/issues/1482)
 - Sketch: It's now possible to disable stroke for polygons and circles [#1177](https://github.com/hajkmap/Hajk/issues/1177)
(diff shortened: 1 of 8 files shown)


sweco-semara:
@Hallbergs Copy/paste bug in CHANGELOG format, my bad - fixed

Hallbergs (approved):
Very nice!

PR #1496, opened 2024-04-09
Feature/1374 configurable search options

### LINKED ISSUE #1374: Make the function "search only in visible layers" optional in admin
**Problem:**
As a user, I expect to get search hits on all activated layers in the map that are within my drawn radius, when I activate the function "search only in visible layers" in the map and search using the radius tool.

For the “search only in visible layers” function to work well, it is therefore preferable that all layers in the map are searchable in this function.

For a layer to be searchable in this function, the wms layers need to be made searchable in the search tool settings in admin and a corresponding wfs layer needs to be available for each wms layer.

It will therefore not be possible for the user to search all visible layers if wfs layers aren’t available that corresponds to all wms layers.

**Proposed solution:**
The proposal is therefore to make the function "search only in visible layers" optional in admin so that we can deactivate this function if wfs layers aren’t available for all wms layers in the map.

linusfj:
Closes #1374.


Hallbergs (approved):
Works well!

The code could be refactored though. For example the following is repeated in several places:

```js
onChange={(e) => {
                const { checked } = e.target;
                this.setState((prevState) => {
                  const enabledSearchOptions = new Set(
                    prevState.enabledSearchOptions
                  );
                  if (checked) {
                    enabledSearchOptions.add("searchInVisibleLayers");
                  } else {
                    enabledSearchOptions.delete("searchInVisibleLayers");
                  }
                  return {
                    enabledSearchOptions: Array.from(enabledSearchOptions),
                  };
                });
              }}

Hallbergs (approved):
I've merged develop and updated the changelog. Tested with Node backend and seems to work fine. @jacobwod @jesade-vbg @Albinahmetaj @OlofSvahnVbg do you want to test it before merge?

Albinahmetaj:
Tested it (with Node backend) and it seems to be working!

Issue #1500, opened 2024-04-12
Cannot create or delete map configurations with Hajk admin since the deletion of API v1


sweco-semara:
**Description**
Cannot create or delete map configurations with Hajk admin since the deletion of API v1.

**To Reproduce**
Use admin to create or delete a map config. This fails always.

**Expected behavior**
Expect latest admin to use API v2, expect maps to be created and deleted as previously possible with API v1.

**Suggested Resolution**
Could be fixed by:
1. Adapting admin mapsettings HTTP metods to expected v2 HTTP methods (which are correctly more REST than v1 modifying GET calls)
1. Adjusting Hajks latest default admin config to expected API v2 mapconfig paths

See PR #1501 for a fix.

PR #1501, opened 2024-04-12
Adapt admin mapconfig endpoints to API v2

sweco-semara:
* Extend API migration documentation with mapconfig steps
* Update Hajk's distributed admin configs
* Resolves hajkmap/Hajk#1500


jacobwod:
Thanks for pointing it out and a quick fix, @sweco-semara!

jacobwod (approved):

Issue #400, opened 2020-01-30
Add info button even for groups (in LayerSwitcher)


jacobwod:
# Background
Currently layers and group layers can have an optional info button visible next to its' name:
![bild](https://user-images.githubusercontent.com/110222/73449620-dace5500-4363-11ea-9564-449d985b649f.png)

Upon click, this info button expands some informational text about the layer/layer group:
![bild](https://user-images.githubusercontent.com/110222/73449640-ec176180-4363-11ea-969c-747bf4b7753d.png)

# Feature request
Users (and admins) have requested that this info button (with corresponding expandable text area) is also available for pure groups (such as Bygga, Bo & Miljö, in the examples above).

PR #1503, opened 2024-04-24
feature: clickable infobutton for groups in layerswitcher

Albinahmetaj:
Issue: #400 

A new feature has been added, similar to the one for layers in the layerswitcher, which allows for writing information about a specific layer; it is  now available for groups.

Please note that I also made some adjustments outside the scope of my affected components. Specifically, I replaced anchor tags with MUI Link for the anchor tags in my current developed group infobutton and for the anchor tags in the layers infobutton (LayerItem.js)


jesade-vbg (approved):
Looks good and tested with good results :)  @OlofSvahnVbg agrees.

PR #1504, opened 2024-04-24
Feature/1494 hajk tooltip component

### LINKED ISSUE #1494: Adding Hajk Tooltip Component to generalise settings and behaviour
The Tooltip messages in SketchView are in the way of a few selection choices:

Skärmbild 2024-04-02 134644

A shown above, for example, the tooltip message for "Mätenhet, areal" is in the way of the first selection ("Automatisk"). The user can not see it and the tooltip message does not go away when going through the choices. The same is true for all three selections:

https://github.com/hajkmap/Hajk/assets/111860819/c25214bd-1d6d-4cf1-98d9-e2023988447e

Additionally, the tooltip messages pop up differently than in Measurer. In "Mät" the tooltip messages seem to be a different component, maybe css styling, and they have a slight delay before appearing:

MÄT

We should make all tooltips be the same mui tooltip AND make them all have a slight delay, which I think is cleaner. We should also move the tooltip messages that are in the way of the selections, or we should make them dissappear when the selection list is opened.

A discussed solution would be a new component "HajkToolTip", which could incorporate a few general stylings and functions. Such as `enterDelay`, `leaveDelay` and other general settings.

OlofSvahnVbg:
closes #1494


apps/client/src/plugins/Measurer/MeasurerView.js
```diff
@@ -25,12 +26,6 @@ const StyledToggleButton = styled(ToggleButton)(({ theme }) => ({
   img: {
     filter: theme.palette.mode === "dark" ? "invert(1)" : "",
   },
-  "&.Mui-selected, &.Mui-selected:hover": {
```
jesade-vbg, line 28: Why was this removed?

jesade-vbg:
I think it looks good and clean! It will be way easier for devs to be consistent. And its not a better feel with the forced default delays.

apps/client/src/plugins/Measurer/MeasurerView.js
OlofSvahnVbg: Think cos we use tooltip in Measurer instead

Hallbergs (approved):
Love it! Feels nice with the new delay.

jesade-vbg (approved):

PR #1506, opened 2024-05-08
Feature/1358 copy coordinates

### LINKED ISSUE #1358: Coordinate: Add a "Copy coordinates" shortcut
Sometimes it's handy to be able to grab the clicked coordinate from Hajk and paste into another application (especially when admin has activated the WGS84 projection which is really interchangeable). Currently it's doable but a somehow cumbersome: user must select each of the fields, copy, paste and put together with a comma (`,`). 
Skärmavbild 2023-06-26 kl  09 23 31

I imagine a small, subtle button next to each of the projection names that will copy a concatenated string (`<N>,<E>`) to clipboard.

OlofSvahnVbg:
closes #1358


Hallbergs:
Netlify fails since the repo structure is updated in develop but not in master. Nothing to worry about. We should set up separate builds for develop and master tbh.

jesade-vbg:
We're going to test some more, so now in Draft mode.

apps/client/src/plugins/Coordinates/CoordinatesTransformRow.js
```diff
@@ -67,6 +81,110 @@ class CoordinatesTransformRow extends React.PureComponent {
     });
   }
 
+  getCoordinates(coordinateFormatTitle) {
+    let inputX, inputY;
```
jesade-vbg, line 85: Sorry but we really need a diffrent solution for this as the coordinate systems are fully customizable in the admin interface.  The can be one coordinatesystem or ten or whatever, and perhaps not SWEREF at all.

apps/client/src/plugins/Coordinates/CoordinatesTransformRow.js
```diff
@@ -67,6 +81,110 @@ class CoordinatesTransformRow extends React.PureComponent {
...
+    // Make the element non-interactive and hide it
+    input.setAttribute("readonly", "");
+    input.style.position = "absolute";
+    input.style.left = "-9999px";
+
+    // We copy the string from the element, alert the user of successful copying...
+    document.body.appendChild(input);
+    input.select();
+    document.execCommand("copy") &&
+      this.props.enqueueSnackbar("Kopiering till urklipp lyckades!", {
```
jesade-vbg, line 133: Maybe:
"Kopierat koordinaten till urklipp"
Opinions?

jesade-vbg (changes requested):

apps/client/src/plugins/Coordinates/CoordinatesTransformRow.js
OlofSvahnVbg: This is correct. The logic for selecting the correct numeric formats x and y value is now updated.

apps/client/src/plugins/Coordinates/CoordinatesTransformRow.js
OlofSvahnVbg: Changed this too. It now says "Koordinaten kopierades till urklipp"

apps/client/src/plugins/Coordinates/CoordinatesTransformRow.js
```diff
@@ -67,6 +80,103 @@ class CoordinatesTransformRow extends React.PureComponent {
...
+    const coordinatesString = `${inputX},${inputY}`;
+
+    // We create a temporary element to store the and copy the coordinateString
+    const input = document.createElement("input");
+    input.value = coordinatesString;
+
+    // Make the element non-interactive and hide it
+    input.setAttribute("readonly", "");
+    input.style.position = "absolute";
+    input.style.left = "-9999px";
```
Hallbergs, line 119: Instead of creating an element to copy from, wouldn't `navigator.clipboard.writeText(coordinatesString)` get the job done? There might be security considerations that i am missing though.

Hallbergs (commented):
Looks good!

Hallbergs (approved):
lgtm

jesade-vbg (approved):

PR #1520, opened 2024-05-29
Feature/1394 documenthandler front page

### LINKED ISSUE #1394: Document Handler - Print Front Page
A few people have asked for a possibility to include a front page when printing documents in, for example, oversiktsplan. The front page should, obviously, come before Content (Innehållsförteckning), which is automatically generated when you print one or more chapters.

A possible solution would be to add options for both Content and Front page:

Perhaps "Include Content" and "Include Front page". The front page could be an image and a title and some other information. These could be taken from existing images and titles depending on the document. Different for FOP and OP and so on...

The problem is deciding when to include what. If you choose "print all documents", should Content and Front page automatically be included. As it is now, Content is always generated, but maybe this should change. Do you need Content if you only want to print one chapter, for instance?

Also, should the possibility for a front page and content be set in admin or should it always be there? There is need for discussion regarding this.

OlofSvahnVbg:
closes #1394


Hallbergs:
What is the status on this one? Still relevant? @OlofSvahnVbg

OlofSvahnVbg:
Status is basically complete but could use some oversight. I can look at it tomorrow.

OlofSvahnVbg:
This is a certified lgtm.

Hallbergs (changes requested):
Unfortunately i don't agree.

image

Also, what is the deal with the lizard?
https://letsenhance.io/static/8f5e523ee6b2479e26ecc91b9c25261e/1015f/MainAfter.jpg

OlofSvahnVbg:
Closing due to inactivity

PR #1525, opened 2024-06-03
Fix to remove multiple animations in Location plugin etc

### LINKED ISSUE #1524: Location plugin - multiple unwanted flash animations.
There is a little bug that causes multiple animations in the Location plugin (Two clicks on the button causes one animation).
There is also a tiny bug which causes the animation not to run the first time (if no rendering occures).
image

jesade-vbg:
Closes #1524


jacobwod (approved):
Thank you 🙏

Issue #1403, opened 2023-09-11
Client UI: Show a confirmation dialog when user attempts to leave the app with unsaved work in the current session


jacobwod:
A couple of users have asked for this now and I can see how important it is for a typical workflow, so let us have a look.

Some of Hajk's tools, such as Sketch, allow users to do a lot of work during a session. (Users often put a substantial amount of time for drawing some features into map and style them properly.) 

But if user happens to close the window or tab with Hajk, all that work will be gone. 

We should prevent it by **adding a warning (or confirmation) dialog to Hajk, saying something like "You have unsaved work in <plugin>, are you sure you want to close the application?"**.

@hajkmap/dev: Up for grabs as I unfortunately don't see myself working on it in the upcoming weeks.

PR #1532, opened 2024-06-25
Feature/1526 measurer select feature

### LINKED ISSUE #1526: Measurer - Add Select Feature
As mentioned in previous issues ([#1311](https://github.com/hajkmap/Hajk/issues/1311)), there should be a select feature in Measurer.
Similar to the one that already exists in Sketch:

image

The function could be accesed via a button that would - preferably - sit in between the circle and delete button:

image

This would enable selecting of a active feature. Clicking this feaure would then automatically measure its circumference using the appropriate measurement (polygon, line, etc.)

OlofSvahnVbg:
closes #1526


jacobwod (commented):
The code looks fine and I was about to approve. However, a quick test (compared with Sketch) reveals that the SelectFeaturesDialog _does not show the highlight_, as Sketch does, making it difficult to see which feature actually gets selected. 

Please see the attached video.

https://github.com/hajkmap/Hajk/assets/110222/f2a0e6ce-c45e-4c8e-bfc7-caef1ff3bb6e

OlofSvahnVbg:
Thanks for the video, will check this

Hallbergs:
> So a few things: SketchModel has this function:
> 
> ```js
> createHighlightFeature = (feature) => {
>     // If no feature (or a feature with no get-geometry) is supplied, we abort.
>     if (feature && feature.getGeometry()) {
>       // Otherwise we create a new feature...
>       const highlightFeature = new Feature({
>         geometry: feature.getGeometry().clone(),
>       });
>       // ...set an id and a highlight-style...
>       highlightFeature.setId(this.generateRandomString());
>       highlightFeature.setStyle(this.#createHighlightStyle());
>       // Finally we return the feature so that we can add it to the map etc.
>       return highlightFeature;
>     }
>   };
> ```
> 
> This is being used via a model passed down to SelectFeaturesDialog in Sketch.
> 
> The problem is that Measurer doesnt have a model. Should we use Sketchmodel in measurer as model to acces this function or what is best practice here?

Maybe the function can be moved to a core model such as the `DrawModel`? We shouldn't share (plugin-)models between plugins (since measurer can be activated while sketch is not).

OlofSvahnVbg:
> > So a few things: SketchModel has this function:
> > ```js
> > createHighlightFeature = (feature) => {
> >     // If no feature (or a feature with no get-geometry) is supplied, we abort.
> >     if (feature && feature.getGeometry()) {
> >       // Otherwise we create a new feature...
> >       const highlightFeature = new Feature({
> >         geometry: feature.getGeometry().clone(),
> >       });
> >       // ...set an id and a highlight-style...
> >       highlightFeature.setId(this.generateRandomString());
> >       highlightFeature.setStyle(this.#createHighlightStyle());
> >       // Finally we return the feature so that we can add it to the map etc.
> >       return highlightFeature;
> >     }
> >   };
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This is being used via a model passed down to SelectFeaturesDialog in Sketch.
> > The problem is that Measurer doesnt have a model. Should we use Sketchmodel in measurer as model to acces this function or what is best practice here?
> 
> Maybe the function can be moved to a core model such as the `DrawModel`? We shouldn't share (plugin-)models between plugins (since measurer can be activated while sketch is not).

Sure. That makes sense

OlofSvahnVbg:
So in DrawModel, we have this prop:

```JSX
  #publishInformation = ({ subject, payLoad }) => {
    // We utilize the fact that this method runs on important changes, such
    // as feature add/remove. We check if there are any features in the draw
    // layer and save that information in the Public API. This is later
    // read in a handler for "onbeforeunload". See #1403.
    if (this.getAllDrawnFeatures().length > 0) {
      window.hajkPublicApi.dirtyLayers[this.#layerName] = true;
    } else {
      delete window.hajkPublicApi.dirtyLayers[this.#layerName];
    }

    // If no observer has been set-up, or if the subject is missing, we abort
    if (!this.#observer || !subject) {
      return;
    }
    // Otherwise we create the prefixed-subject to send. (The drawModel might have
    // been initiated with a prefix that should be added on all subjects).
    const prefixedSubject = this.#observerPrefix
      ? `${this.#observerPrefix}.${subject}`
      : subject;
    // Then we publish the event!
    this.#observer.publish(prefixedSubject, payLoad);
  };
```

Why do we need this?

Hallbergs:
> So in DrawModel, we have this prop:
> 
> ```js
>   #publishInformation = ({ subject, payLoad }) => {
>     // We utilize the fact that this method runs on important changes, such
>     // as feature add/remove. We check if there are any features in the draw
>     // layer and save that information in the Public API. This is later
>     // read in a handler for "onbeforeunload". See #1403.
>     if (this.getAllDrawnFeatures().length > 0) {
>       window.hajkPublicApi.dirtyLayers[this.#layerName] = true;
>     } else {
>       delete window.hajkPublicApi.dirtyLayers[this.#layerName];
>     }
> 
>     // If no observer has been set-up, or if the subject is missing, we abort
>     if (!this.#observer || !subject) {
>       return;
>     }
>     // Otherwise we create the prefixed-subject to send. (The drawModel might have
>     // been initiated with a prefix that should be added on all subjects).
>     const prefixedSubject = this.#observerPrefix
>       ? `${this.#observerPrefix}.${subject}`
>       : subject;
>     // Then we publish the event!
>     this.#observer.publish(prefixedSubject, payLoad);
>   };
> ```
> 
> Why do we need the prefixedSubject?

If i remember correctly, we use the prefix so that plugins etc. that uses the `DrawModel` can catch "their" events. For example: the measure plugin might initiate a `DrawModel` with "measure-draw-model".

jacobwod (dismissed):
Seems to work rather nice now. You have _one tiny_ warning to fix, then it's ready to merge. Feel free to add a CHANGELOG row about it too!

```
[eslint] 
src/plugins/Sketch/models/SketchModel.js
  Line 3:32:  'Style' is defined but never used  no-unused-vars
```

OlofSvahnVbg:
Fixed warning and added to changelog

jacobwod (approved):

PR #1546, opened 2024-09-18
Feature/1545 overlap fixes

### LINKED ISSUE #1545: Search: Overlap fixes
There is some issue with the overlap of the search window in app:

image

As shown above, the search result window under the search bar is always displayed on top, which is not ideal when working with one or more control windows placed to the right of the screen.

This has been reported by some users as a problem.

The drawerButtons have a similiar problem when overlapping with other windows:

image

Otherwise, the control windows behave correctly when interacting with each other. The active window is put on top when clicked or interacted with, as it should.

I think this should be fixed by changing logic for search and drawerbuttons.

OlofSvahnVbg:
closes #1545

### DIFF
--- a/apps/client/src/components/App.js
+++ b/apps/client/src/components/App.js
@@ -487,6 +487,9 @@ class App extends React.PureComponent {
       .loadPlugins(this.props.activeTools);
 
     Promise.all(promises).then(() => {
+      this.globalObserver.subscribe("handleBlur", () => {
+        this.setState({ headerHasFocus: false });
+      });
       // Track the page view
       this.globalObserver.publish("analytics.trackPageView");
 
@@ -698,6 +701,7 @@ class App extends React.PureComponent {
 
     // Register various global listeners.
     this.globalObserver.subscribe("infoClick.mapClick", (results) => {
+      console.log(results);
       this.setState({
         mapClickDataResult: results,
       });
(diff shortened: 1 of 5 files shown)


jacobwod:
I'd love to start using it, but unfortunately there are some issues to be resolved. Does anyone else see the same behaviour?

https://github.com/user-attachments/assets/8dcaa7b3-c4c0-4c12-ae38-df2b61607116

Hallbergs:
I have the same behavior as you @jacobwod

Albinahmetaj:
I checked out the branch and read the issue description for clarification to myself. However I do not get the same "flickering" as you showed in the video above @jacobwod, am I missing something? I tested the branch as it is and did not apply any extra configuration.

jacobwod:
The flickering appears to be a Safari-only issue. 

Would be interesting to try it out on iOS Safari too, haven't got the opportunity though.

OlofSvahnVbg:
So, should we accept this as is and handle the flickering issue as a separate bug?

jesade-vbg:
> So, should we accept this as is and handle the flickering issue as a separate bug?

Sorry but no, in my opinion. Maybe if we had a working fix brewing, but we don't right now.

Hallbergs:
@jesade-vbg @OlofSvahnVbg If i remember correctly we agreed that @OlofSvahnVbg was to create a separate issue since the flickering does not seem to be related to this PR. (We have flickering when using safari in develop as well).

@OlofSvahnVbg please create a separate issue so that we can merge this PR.

Hallbergs (approved):

PR #1549, opened 2024-11-05
Measurement segments

michaelsund:
Added segment-measurements of linestrings and polygons to the measurer plugin, and extended some geometry removal functionality in DrawModel.

Should maybe have a admin-option to enable disable segment measurements all together, and hide the button.

image

image

### DIFF
--- a/apps/client/src/plugins/Measurer/Measurer.js
+++ b/apps/client/src/plugins/Measurer/Measurer.js
@@ -266,13 +266,11 @@ function Measurer(props) {
     restoreHoveredFeature();
     angleSnapping.setActive(false);
     angleSnapping.clearSnapGuides();
-    // segments.setEnabled(false);
     setPluginShown(false);
   };
 
   const onWindowShow = () => {
     angleSnapping.setActive(true);
-    // segments.setEnabled(true);
     setPluginShown(true);
   };
 
--- a/apps/client/src/plugins/Measurer/Measurer.js
+++ b/apps/client/src/plugins/Measurer/Measurer.js
@@ -12,11 +12,13 @@ import DrawModel from "../../models/DrawModel";
 import { Circle, Fill, RegularShape, Stroke, Style } from "ol/style";
 import HelpIcon from "@mui/icons-material/Help";
 import AngleSnapping from "./AngleSnapping";
+import Segment from "./Segment";
 
 function Measurer(props) {
   const { map, app } = props;
   const [state] = React.useState({});
   const currentHoverFeature = useRef(null);
+  const [segmentsEnabled, setSegmentsEnabled] = useState(false);
   const [localObserver] = React.useState(Observer());
   const [drawType, setDrawType] = useState("LineString"); // default to LineString as previous Measure tool
   const [pluginShown, setPluginShown] = React.useState(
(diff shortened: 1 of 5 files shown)


apps/client/src/plugins/Measurer/Measurer.js
```diff
@@ -245,11 +266,13 @@ function Measurer(props) {
     restoreHoveredFeature();
     angleSnapping.setActive(false);
     angleSnapping.clearSnapGuides();
+    // segments.setEnabled(false);
```
Hallbergs, line 269: Remove unused code

apps/client/src/plugins/Measurer/Measurer.js
```diff
@@ -245,11 +266,13 @@ function Measurer(props) {
...
     setPluginShown(false);
   };
 
   const onWindowShow = () => {
     angleSnapping.setActive(true);
+    // segments.setEnabled(true);
```
Hallbergs, line 275: Remove unused code

apps/client/src/plugins/Measurer/Segment.js
```diff
@@ -0,0 +1,289 @@
...
+  #currentSegmentLength;
+  #polygonFirstAndLast;
+  #currentCenter;
+  #measureFeature;
+  #map;
+  #currentFeatureSavedId;
+  #currentPoint;
+
+  constructor(drawModel, map, enabled) {
+    this.#enabled = enabled | false;
```
Hallbergs, line 17: Interesting with the bitwise OR here. Should get the job done. However, the nullish coalescing operator is common in Hajk, so let's stick to that to avoid confusion.

```js
this.#enabled = enabled ?? false;
```

Hallbergs (changes requested):
Well commented and structured code! Works perfectly! Nice!

Just a couple of small fixes (see comments). Also, there is a small gap between the old buttons and the new one. Intentional?

Good job!

Hallbergs:
Also, please take a look at:
image

michaelsund:
Hi! Pushed the mentioned fixes, thanks for the feedback!

> Also, there is a small gap between the old buttons and the new one. Intentional?
Yes, I created the new buttongroup with spacing on purpose to indicate that it's not a drawing tool but an "option", but there's probably even better ways to make that clear with different placement or other toggle styles.

jacobwod:
Regarding the button that activates/deactivates showing segment measurements: we don't have any specific settings toolbar or window in the Measure tool at this time. Adding it for just one setting is an overkill, in my opinion. So your solution (a separate button outside the tools button group) is good. 

However, please consider hiding this button by default and only showing it when line measuring is activated (as I assume it's only relevant for lines?).

Hallbergs:
> Regarding the button that activates/deactivates showing segment measurements: we don't have any specific settings toolbar or window in the Measure tool at this time. Adding it for just one setting is an overkill, in my opinion. So your solution (a separate button outside the tools button group) is good.
> 
> However, please consider hiding this button by default and only showing it when line measuring is activated (as I assume it's only relevant for lines?).

It's useable for both lines and polygons. However, it could still be a good idea to hide it when it's not useable.

Hallbergs (approved):
Looks good!

I think it's a nice addition even if the button is showed at all times, so i'll approve the PR and hold off the merge for a couple of days. LMK if you plan to implement the hiding feature or not :)

michaelsund:
Nice! I'll take a look at it this weekend.

Hallbergs (approved):
Very good! Thank you for contributing!

PR #1552, opened 2024-11-18
Fix/1547 bookmarks sublayers not displayed

### LINKED ISSUE #1547: Sublayers not displayed in Bookmarks plugin
**Describe the bug**
When activating a bookmark that contains group layers from WMS services, the sublayers within the group layers are not properly displayed on the map. The group layer itself is activated, but its sublayers are not visible.

**To Reproduce**

1. Add a WMS group layer that contains sublayers to the map.
2. Create and save a bookmark that includes the group layer.
3. Activate the bookmark.
4. Notice that while the group layer appears as activated, its sublayers are not visible on the map.

**Expected behavior**
When a bookmark containing a WMS group layer is activated, the sublayers should also be displayed correctly on the map.

linusfj:
Closes #1547.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Search: Added possibility to set default search options via admin UI. PR: [#1496](https://github.com/hajkmap/Hajk/pull/1496)
+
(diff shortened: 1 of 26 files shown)


Hallbergs:
I see that there are some changes in the LayerSwitcher included. Have you discussed this with with Jakob who is working on the new LayerSwitcher?

Hallbergs:
Closing due to #1568

PR #1568, opened 2024-11-27
LayerSwitcher: New take on Filtering, QuickAccess, QuickAccessPresets and DrawOrder improvements

karljakoblarsson:
# A Improved Layer Switcher.

This PR is a continuation of the work done by @dennisbergstrom and @linusfj.  Based on the issue: https://github.com/hajkmap/Hajk/issues/1380
A previous version of this change was proposed in: https://github.com/hajkmap/Hajk/pull/1437 However that PR did not meet the projects standards.

### The main changes since #1437 are:

- Merge with latest develop
- Several Performance improvements
- Bug fixes
    - Including a fix so the window scrolls both for `height: "auto"` and `"dynamic"`
- Easier to read code
- Match styles to previous version

Please help me and test this with different configs. I have mostly tested with SimpleMapConfig, Halmstads public and Göteborgs internal config.

### Current status of this PR:

This PR is not intended as the final state for the LayerSwitcher.  The current code is not in a state I would be willing to support if I were the maintainer of this project.  But I think the new features still makes this PR a net improvement compared to current `develop`.  I will keep working on the LayerSwitcher for at least a few weeks.

The most important work still to be done is to make the code simpler and easier to read.  That will hopefully also improve the performance, or at least make it easier to use to improve it in the future.  There is also several improvements and missing features from the original plan to be done.

#### Features Missing from this PR:
- Editor for the QuickLayersPresets (Layer themes/packages) in the admin UI.
    We talked about this at the "Kodgrupp"-meeting today  (2024-11-27) that we will implement that in the new admin interface in a separate PR. It would be wasteful to implement it in the old admin interface.

#### Still needs more testing:
- CQL-filtering for vector layers.

### Refactorings to be done/in progress:
- Use functional components instead of class based in most cases.
- Remove the use of `this.forceUpdate()` and similar hacks to re-render components when OpenLayers state changes but props/state does not.
- Refactor map state management so it's only very few componenent which modify the OpenLayers state/objects
- Simplify data flow such that components only get the data they need.
- Simplify DOM and Component tree
- Refactor QuickAccessPresets and make sure naming is consistent

### Features/improvements I would like to add in a future change
- Don't show QuickAccess toggle icon if section is empty
- Save QuickAccess layers in LocalStorage
- Open groups when filtering layers so it's possible to see the hits.
- Harmonize the Layer/Group info display
- Make the drag&drop skeleton look properly
- Keyboard navigation in the LayerSwitcher
- When applying a Layer package/theme open the QuickAccess
[...]

PR #1574, opened 2024-12-20
Refactor LayerSwitcher

karljakoblarsson:
This is a refactoring of the LayerSwitcherPlugin. A continuation of the work started in #1568.

The goal of this PR is to simplify the code in the LayerSwitcher, so it will be easier to work on in the future. To align it more with the design principles of React. To remove most of the event subscriptions in child components and centralize all state. OpenLayers should be the source-of-truth for everything map-related. Most components should be unaware of OpenLayers specifics. I also tried to decrease the amount of prop-drilling.

Another goal was to reduce the amount of code quite a lot. I didn't not reach that goal in this PR, but I think the added lines are much easier to read than the removed lines.

The performance should not be much worse. Since I moved a lot of state-handling code it will be different. Some renders might take a little longer, but most should be faster. But I need help to test this.

Something is weird with the commit history in this PR where all commits from #1568 is shown as well. But the "Files Changed" is correct.

Added features:
---------------
- When searching/filtering: Every LayerGroup and GroupLayer with a search it in a child is opened. Even for sublayers.
- When searching/filtering: The QuickAccess is hidden.
- The QuickAccess "expand-arrow" is hidden until there are layers added.
- The warnings about Zoom min/max are bugfixed and moved to a central location.  I don't think that feature belong in the LayerSwitcher long term. It should be easier to move now.

General "architecture"
----------------------

The file `LayerSwitcherProvider.js` is responsible for all state and the synchronization between React and OpenLayers. It listens for changes in OpenLayers and updates it's state based on that.

It also provides the `useLayerSwitcherDispatch` hook. The hook returns a object where a component can initiate an "action", like "Toggle layer with ID". This means that the individual components does not need to know the specifics on how to toggle a layer.

There is also the Zoom min/max check. That is a separate component who's only job is to send the toast warning.


Hallbergs (changes requested):
Great job! I've tested it locally with our config and I've found a couple of things:

1. Very nice that the groups expand when filtering the layers! However, when i use a config with around 500 layers i experience severe lag (hangs for 3-5 sec on first character. This is on dev environment so it won't be that bad in prod!). I suggest that we limit filtering to input larger than around 3 chars?
2. I can see my legend in the "list view", but not under the detail view. Could be related to some missing config-setting?
image
image

Also, there's a commended `console.log` in app.js :)

Hallbergs:
Minimum filter lenght as a config setting sounds great!

karljakoblarsson:
@Hallbergs 
I've added two commits that does three things:

1. Removes the eslint warnings and the console.log
2. Fixes the Legend URL in the layer details view so that it shows up as it should
3. Adds a 3 character minimum for the layer search/filtering

The minimum for the layer filtering work that the filter is only applied when the text is 3 or more characters, or if the user presses enter. If the filter text is fewer characters a helper text is shown below the InputField. (The styles could be improved, and in the future a delay could be added before the message show up, if we keep this solution.) The minimum filter text length can be set by the setting `LayerSwitcherOptions.layerFilterMinLength`, the default is 3.

karljakoblarsson:
https://github.com/user-attachments/assets/0350db0e-5949-46b9-9ef5-82c132e7b18b

karljakoblarsson:
https://github.com/user-attachments/assets/482924b8-e762-4e40-b48b-57a606760771

I'm starting to think that the tooltip is a better UI

Hallbergs:
Very good! I think i like the tooltip better as well. I'll test it again early next week!

karljakoblarsson:
I changed the help text when `< min filtertext lenght` to a tooltip and added a bug fix I found for QuickAccess

karljakoblarsson:
@Hallbergs Found a few bugs, mostly about VectorLayers. (Which I apparently didn't test enough)

I pushed a few commits fixing those and some other minor things I found.
```
[fix: filter/serach for nested groups](https://github.com/hajkmap/Hajk/pull/1574/commits/423e72d85565e3c2c95f00dfdbb3f414c96d2580)
[423e72d](https://github.com/hajkmap/Hajk/pull/1574/commits/423e72d85565e3c2c95f00dfdbb3f414c96d2580)
[fix: bug when changing bg without OSM](https://github.com/hajkmap/Hajk/pull/1574/commits/9ce65063a908ebc6bfe3c7cbc6295a99c05b41a1)
[9ce6506](https://github.com/hajkmap/Hajk/pull/1574/commits/9ce65063a908ebc6bfe3c7cbc6295a99c05b41a1)
[fix: toggle VectorLayers without crashing](https://github.com/hajkmap/Hajk/pull/1574/commits/727b29356e6d17420c5f5a6a79b9b7128d2223a3)
[727b293](https://github.com/hajkmap/Hajk/pull/1574/commits/727b29356e6d17420c5f5a6a79b9b7128d2223a3)
[fix: don't crash with VectorLayers in QuickAccess](https://github.com/hajkmap/Hajk/pull/1574/commits/75685fc76ccc3edad7268be607be871cba6bda6e)
[75685fc](https://github.com/hajkmap/Hajk/pull/1574/commits/75685fc76ccc3edad7268be607be871cba6bda6e)
[fix: don't crash for VectorLayers in DrawOrder](https://github.com/hajkmap/Hajk/pull/1574/commits/14b17f53557d5f5d734871146bc876023d5ff387)
[14b17f5](https://github.com/hajkmap/Hajk/pull/1574/commits/14b17f53557d5f5d734871146bc876023d5ff387)
[fix: remove not visible quick access layers](https://github.com/hajkmap/Hajk/pull/1574/commits/5f9f25315ed7a2fe3ee82ea8873a714e0c65b2d7)
```

karljakoblarsson:
I found that the "min length" search tooltip didn't read the config setting correctly. So pushed a small fix for that.

Albinahmetaj:
I tested with my config yesterday before the meeting and it works well!

Hallbergs (approved):

Hallbergs:
I'll hold off with the merge until tomorrow, just in case someone else wants to test it!

Issue #1580, opened 2025-01-20
LayerSwitcher: new LayerSwitcher, filtering for layers in a group where some layers match the filter returns erroneus visibility of layers and info.


eri-blan:
**Describe the bug**
In develop from 2025-01-17, If a search filter is made where the filter matches certain layers and not others, all layers in the layer group will be  set visible if clicking on one of them, even layers that don´t match the filter. Also, the info button returns information for the wrong sublayer.

Here is a short video showing the issue:
[Clip](https://github.com/user-attachments/assets/432f5049-965b-49d5-9d04-e7fcbe8b331b)

**Expected behavior**
Only clicked layers will be set to visible/not visible. Info-button returns the correct sublayer information.

Hallbergs:
Good find. @karljakoblarsson would it be possible for you to have a look at it?

karljakoblarsson:
For me the bug seems to be:

The first click on a sublayer on a non-visible grouplayer toggles the whole grouplayer and not just the sublayer. I don't think there is any connection to the filtering.

Is that correct @eri-blan ?

karljakoblarsson:
Fixes here: https://github.com/hajkmap/Hajk/pull/1581

eri-blan:
> For me the bug seems to be:
> 
> The first click on a sublayer on a non-visible grouplayer toggles the whole grouplayer and not just the sublayer. I don't think there is any connection to the filtering.
> 
> Is that correct [@eri-blan](https://github.com/eri-blan) ?

Yes, I think that is correct regarding the layer visibility. The information part might have a connection to the filtering, but not 100% sure.

eri-blan:
> Fixes here: [#1581](https://github.com/hajkmap/Hajk/pull/1581)

Great, thanks! I´ll try again when it is merged 👍

Hallbergs:
@eri-blan now it is merged and ready to test!

PR #1581, opened 2025-01-20
Fix for "LayerSwitcher: new LayerSwitcher, filtering for layers in a group where some layers match the filter returns erroneus visibility of layers and info."

karljakoblarsson:
Bug fixes for #1580 See the issue for details.


Hallbergs (approved):
Did some quick tests and it seems to fix the issue!

PR #1588, opened 2025-02-11
Fix: Handle missing layer config in LayerSwitcher

karljakoblarsson:
@jacobwod found a bug in the LayerSwitcher to other day.

If a layer is configured in the layer switcher settings, but is missing or misconfigured in the `layersConfig` the LayerSwitcher crashed.

The desired behavior is that we ignore any layer that is missing from the layers config. This PR adds that check to the LayerSwitcher initialization. And also makes the `LayerItem` component a bit more robust agains missing data.

### DIFF
--- a/apps/client/src/plugins/LayerSwitcher/LayerSwitcherProvider.js
+++ b/apps/client/src/plugins/LayerSwitcher/LayerSwitcherProvider.js
@@ -307,6 +307,17 @@ const getLayerNodes = (groups, olLayerMap) =>
 
     const olLayer = olLayerMap[node.id];
 
+    // If we're looking at a "normal layer" (not a group) and the OpenLayers
+    // object is missing that means that the layer configured in the
+    // LayerSwitcher options but is missing or misconfigured in the
+    // `layersConfig`. In that case we should ignore the layer and don't show
+    // it in the LayerSwitcher
+    if (!isGroup && olLayer === undefined) {
+      // A non group layer should not have any children so it's okay to return
+      // the empty array here.
+      return [];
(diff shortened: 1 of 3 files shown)


Hallbergs (changes requested):
Unfortunately `simpleMapAndLayersConfig` crasches. See below:

jacobwod (changes requested):
So the crash is surely fixed with our config, but also there are no layers that pass through the new filter… 😄 So it's kind of empty.  (Also for background layers btw.)

 Skärmavbild 2025-02-12 kl  08 45 07

karljakoblarsson:
@jacobwod @Hallbergs Do you want to quickly test the last commit I did right now. It might be that easy to fix.

Hallbergs:
The simple config works now!

jacobwod (changes requested):
Hmm still issues here. 😢 

The main thing seems to be:
> `undefined is not an object (evaluating 'olLayer.setVisible')`

which throws in LayerSwitcherProvider on line 239. 

Only this time, I can expand my layers, which seems a bit unexpected (they are treated as Group Layers):
Skärmavbild 2025-02-12 kl  11 01 02

The crash does not occur on load nor on expand, but rather once I check the visibility checkbox for any (group)layer.

karljakoblarsson:
@jacobwod Is it possible to share some configuration? Or some version at least that still has the problem?

jacobwod (approved):
I conducted further extensive testing and could not reproduce the earlier crash. Sorry for the delay, good job! 🚀

PR #1589, opened 2025-02-11
Fix: Disable LayerItemDetails for system layers in DrawOrder tab

karljakoblarsson:
@Albinahmetaj found a bug the other day.

The LayerItemDetails view crashed for system layers. (Which can be shown in the DrawOrder tab)

This PR fixes that issues in two ways:

- It improves the error handling in LayerItemDetails such that it does not crash. But it does not really show any useful info for most system layers.
- It disables the arrow button to open the LayerItemDetails for system layers.


jacobwod (approved):
Good fix!

Hallbergs (approved):

Issue #1586, opened 2025-01-31
layerSwitcher: new layerSwitcher, question regarding toggling of visibility for group layers when search filter is active


eri-blan:
We noticed that when a search filter is active such that some sublayers are visible in the layerSwitcher, toggling visibility for the parent group toggles visibility for all sublayers in the group, even those not shown in the layerSwitcher when the search filter is active. It might be confusing for the user but maybe it is by design. Any thoughts on changing this? Perhaps so that only the filtered layers in the group get toggled.

karljakoblarsson:
@Hallbergs @jacobwod do you have any opinions about this? I haven't used Hajk enough to feel like I can have an opinion.

The code change wouldn't be that complicated I think.

Hallbergs:
I think that only toggling the layers that are currently shown would be the most intuitive solution. If it is a somewhat simple change i say go! @karljakoblarsson 

If you implement it, please put the PR against master. I think we can see this as a bugfix.

karljakoblarsson:
@eri-blan @Hallbergs 

Another question:

If a whole group is toggled.
Then I filter such that only a few layers are visible in the layerSwitcher.
And then presses untoggle on the group.
Should only the filtered layers be hidden then?

jacobwod:
Hmm, on the other hand: what I propose above would be inconsistent with this issue's proposal of only enabling visible layers.

OK, so what's in #1599 currently? Perhaps the best way to answer it would be to try things out.

karljakoblarsson:
We discussed this at the code group meeting this morning. Since we couldn't agree on the correct behaviour we decided to keep the current behaviour for the 4.0 release. (Toggling a group toggles the entire group on or off.) There are some open questions about how groups and layers with sublayers should work.

There is a plan to expand the search to include descriptions and keywords in a future version. Then there is the need to highlight the actual search hit and where it is. That would be a good timing to revisit this.

PR #1599, opened 2025-02-24
feat: Toggle only filtered layers in Groups

karljakoblarsson:
This PR is a implementation/fix for #1586 

If there is a active filter/search in the LayerSwitcher and the user enables a Group of GroupLayer we only toggles the layers which are search hits.


karljakoblarsson:
@Hallbergs @jacobwod I have a question about the functionality:

> - A whole group is toggled.
> - Then filter such that only a few layers are visible in the LayerSwitcher.
> - And then presses untoggle/hide on the group.
> - Should only the filtered layers be hidden then? Or should all be hidden?

Should it be that way for both Groups and GroupLayers (with sublayers)? Or different.

PR #1601, opened 2025-02-24
Feature: List layers in Quick access presets infobox

karljakoblarsson:
Screenshot 2025-02-24 at 11 47 37

Show the layers in the Quick Access Preset in the Infobox. 

@jacobwod had this idea in: https://github.com/hajkmap/Hajk/discussions/1596#discussioncomment-12276132

Also add a example QuickAccess Preset to the default map config.

### DIFF
--- a/apps/backend/App_Data/map_1.json
+++ b/apps/backend/App_Data/map_1.json
@@ -665,7 +665,36 @@
             "infobox": ""
           }
         ],
-        "quickLayersPresets": [],
+        "enableQuickAccessPresets": true,
+        "quickAccessPresets": [
+          {
+            "id": "sensate-data",
+            "title": "Senaste data",
+            "author": "Hajks kodgrupp",
+            "Description": "Det här är en grupp av den senaste datan där vi har flera olika år.",
+            "keywords": ["naturvårdsverket", "2019", "scb", "2023" ],
+            "layers": [
+              {
+                "id": "1",
+                "drawOrder": 3,
+                "visible": true,
+                "opacity": 1
+              },
+              {
+                "id": "toimzo",
+                "drawOrder": 3,
+                "visible": true,
+                "opacity": 1
+              },
+              {
+                "id": "vzjhpz",
+                "drawOrder": 3,
+                "visible": true,
+                "opacity": 1
+              }
+            ]
+          }
+        ],
         "active": true,
         "visibleAtStart": false,
         "visibleAtStartMobile": false,
--- a/apps/backend/App_Data/map_1.json
+++ b/apps/backend/App_Data/map_1.json
@@ -688,9 +688,7 @@
               },
               {
                 "id": "vzjhpz",
-                "drawOrder": 3,
-                "visible": true,
-                "opacity": 1
+                "visible": true
               }
             ]
           }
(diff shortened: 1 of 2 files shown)


jacobwod:
I'm not sure it's working (unless I'm testing it the wrong way).

Here's the relevant portion of my LocalStorage object:

```json
"quickAccessLayers": [
        "1098",
        "rycfnb",
        "948"
    ],
    "layerswitcher": {
        "savedLayers": [
            {
                "layers": [
                    {
                        "id": "1098",
                        "visible": true,
                        "subLayers": [],
                        "opacity": 1,
                        "drawOrder": 133
                    },
                    {
                        "id": "rycfnb",
                        "visible": true,
                        "subLayers": [
                            "befolkning_tatort_1980",
                            "befolkning_tatort_1990",
                            "befolkning_tatort_1995",
                            "befolkning_tatort_2000",
                            "befolkning_tatort_2005",
                            "befolkning_tatort_2010",
                            "befolkning_tatort_2015",
                            "befolkning_tatort_2018",
                            "befolkning_tatort_2020"
                        ],
                        "opacity": 1,
                        "drawOrder": 89
                    },
                    {
                        "id": "948",
                        "visible": true,
                        "subLayers": [],
                        "opacity": 1,
                        "drawOrder": 29
                    },
                    {
                        "id": "774",
                        "visible": true,
                        "subLayers": [],
                        "opacity": 1,
                        "drawOrder": 0
                    }
                ],
                "metadata": {
                    "savedAt": "2025-02-26T06:59:31.724Z",
                    "numberOfLayers": 4,
                    "title": "3 lager",
                    "description": "Lite testlager",
                    "mapName": "Standardkartan"
                }
            }
        ]
    }
```

As you can see, there are 3 layers in quick access (haven't been removed yet, but it's not really relevant here), plus 1 background layer. This sums up to the `numberOfLayers` which is correct, 4.

However, I only see this:
Skärmavbild 2025-02-26 kl  08 03 46

---

An additional problem (which should perhaps be handled in a separate issue?) is the group layer (id `rycfnb`) does not show its children in the Quick Access portion (but it works well in the regular group, see the screenshot):
image

karljakoblarsson:
I misunderstood. This PR adds it for the "presets" from the config. Not the "user saved preset". I'll add it for that dialog as well.

karljakoblarsson:
> An additional problem (which should perhaps be handled in a separate issue?) is the group layer (id rycfnb) does not show its children in the Quick Access portion (but it works well in the regular group, see the screenshot):

Yes, thats a bug! And let's consider it a separate issue.

karljakoblarsson:
Now I added the same list in the "QuickAccess favorites".

However, the code in this PR really needs a big refactoring. And the design could be better.

Since this is an additional feature that was not part of the original plan. So maybe this should be done for a future version and not 4.0. How do you feel about that @jacobwod @Hallbergs

Hallbergs:
Since it needs refactoring and potential design updates i don't think we should include it in the 4.0 version. I'll leave the last call for @jacobwod

jacobwod:
I feel that if it works in its current state, let's merge it. Refactoring can be done in `develop` and merged upstream to `master` in a future release.

Hallbergs:
I think that there are some layout changes needed that can be seen when having many layers:
image

karljakoblarsson:
> I think that there are some layout changes needed that can be seen when having many layers:

Yep, the design needs some work. I normal bullet list should work better in most cases I think. But I will prfioritize bugs before this.

PR #1614, opened 2025-03-06
fix: add map rotation

karljakoblarsson:
https://github.com/hajkmap/Hajk/discussions/1596#discussioncomment-12412108

The new LayerSwitcher accidentally removed the feature to rotate the map when a layer is activated. If the `rotateMap` property is set. This PR re-adds that feature.

@jacobwod I have a few questions about the details in this feature:

- Should the map reset to north is a layer without the `rotateMap` setting. Or should it keep in the current rotation.
- Is this feature only for base layers or any layer?

### DIFF
--- a/apps/client/src/components/App.js
+++ b/apps/client/src/components/App.js
@@ -6,6 +6,7 @@ import PropTypes from "prop-types";
 import { styled } from "@mui/material/styles";
 import Observer from "react-event-observer";
 import { isMobile } from "../utils/IsMobile";
+import { mapDirectionToAngle } from "../utils/mapDirectionToAngle";
 import { getMergedSearchAndHashParams } from "../utils/getMergedSearchAndHashParams";
 import SrShortcuts from "../components/SrShortcuts/SrShortcuts";
 import Analytics from "../models/Analytics";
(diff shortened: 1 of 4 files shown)


jacobwod:
Really good questions @karljakoblarsson!

The way I've seen it being used is solely for background layers, meaning that we don't want to reset it if user enables some other layer (e.g. property borders). The orientation should the way the active background layer says it should be. This way it works for that use case.

But sure, there may be other use cases where the requirements would vary. 

In my opinion (and that's just an opinion) we should stick to only rotating on background layers, for now.

This way, we can ignore your first question (regarding resetting), as there's only one active background layer at a time and we should use its requested orientation.

Is that the way you've implemented it in this PR btw?

karljakoblarsson:
I think you're right in that this makes the most sense for base layers. I tried it with not reseting the rotation, and that felt very weird.

So I changed it such that it only affects base layers.
And if the `rotateMap` setting isn't set we default to rotating the layer to north.

And unfortunately I need to add a hacky workaround for the OpenStreetMap Layer if it's enabled.

jacobwod (approved):
Well done 👍

PR #1617, opened 2025-03-12
fix: hide sublayers when searching hideExpandArrow

karljakoblarsson:
@jesade-vbg found a bug in the layer switcher

> When searching/filtering "hideExpandArrow": true is ignored showing "hidden" layers.

https://github.com/hajkmap/Hajk/discussions/1596#discussioncomment-12423761

This PR fixes that bug.


Hallbergs (approved):
GJ!

PR #1625, opened 2025-03-18
fix: set layer ids from 'ladda tema' to local storage, toggle subLayers visibility

Albinahmetaj:
I sat with @Hallbergs and discussed some cases regarding this commit. My guess would be that if a user decides to load a 'tema' , the layers that are stored in 'Snabbåtkomst' after the 'tema' has been loaded should be stored in local storage as they currently do when a user uses the 'Lägg till tända lager' function.

Another fix ensures that when loading a 'tema' using a grouplayer the specified sublayers will be toggled.

---

Bug: right now all sublayers within the grouplayer will be toggeled when loading layers using  'Ladda tema', even if only two of them are specified  inside the quickAccessPresets as shown _here:_

`  "quickAccessPresets": [layer:[{ "id": "n0ehiv",
                "visible": true,
                "subLayers": [
                  "Motortrafikled",
                  "Gastvag"
                ],
                "opacity": 1,
                "drawOrder": 1000}]]
`

tema grplayer-fix

`olLayer.set("subLayers", visibleSubLayersArray);` fixes this inside the LayerSwitcherProvider.js component

Another thing that i notices is that when you load the layers using 'Ladda tema' with a grouplayer, all sublayers are toggled as described above. However, if you manually untoggle the sublayers and then load the 'tema' again, the sublayers remain unchecked, unlike the first time.

Can you take a look at this @karljakoblarsson and test if this solves the issues

### DIFF
--- a/apps/client/src/plugins/LayerSwitcher/LayerSwitcherProvider.js
+++ b/apps/client/src/plugins/LayerSwitcher/LayerSwitcherProvider.js
@@ -13,6 +13,9 @@ import { useLayerZoomWarningSnackbar } from "./useLayerZoomWarningSnackbar";
 import { functionalOk as functionalCookieOk } from "../../models/Cookie";
 import LocalStorageHelper from "../../utils/LocalStorageHelper";
 
+export const QUICK_ACCESS_KEY = "quickAccess";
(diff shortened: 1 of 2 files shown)


Hallbergs:
@karljakoblarsson If you have time it would be great if you could do a sanity check on this one!

PR #1626, opened 2025-03-19
Fix/documenthandler modal update

OlofSvahnVbg:
fixes (hopefully) the issue with table of content manipulation per document in admin. [Link to conversation](https://github.com/hajkmap/Hajk/discussions/1596#discussioncomment-12427455)


Hallbergs (approved):
I've tested it and it seems to work! Would be grateful if someone who has used the documenteditor more would test it as well!

PR #1633, opened 2025-04-01
Feat/hide layer legend setting in client

karljakoblarsson:
This PR adds the setting `"hideLayerLegend": true,` on layers in the LayerSwitcher settings. It makes it possible to disable the legend for specific layers. This is a request from Göteborg Stad. The setting needs to be added in the Admin tool as well.


Hallbergs:
Closing for now. Let's create a new PR against develop if this one is still valid.

PR #1635, opened 2025-04-03
fix: enable hash control over individual sublayers

karljakoblarsson:
@Hallbergs found some wrong looking code. And it turns out there is an actual bug there. External control over the hash parametes would not control group layers correctly.

This PR should fix that bug.

Thanks to @jacobwod for the`http://localhost:3000/examples/embedded.html` tool that made it so much easier to test this.

### DIFF
--- a/apps/client/src/models/AppModel.js
+++ b/apps/client/src/models/AppModel.js
@@ -38,7 +38,7 @@ import VectorLayer from "ol/layer/Vector";
 import VectorSource from "ol/source/Vector";
 import { Icon, Fill, Stroke, Style } from "ol/style";
 
-import { setOLSubLayers } from "../plugins/LayerSwitcher/LayerSwitcherProvider.js";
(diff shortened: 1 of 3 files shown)


jesade-vbg:
I can see that AppModel now is dependent on LayerSwitcher, which is a plugin. But... LayerSwitcher isnt really a real plugin so maybe its ok this time.

karljakoblarsson:
> I can see that AppModel now is dependent on LayerSwitcher, which is a plugin. But... LayerSwitcher isnt really a real plugin so maybe its ok this time.

Yeah that's bad. The layer switcher currently does a whole lot of layer actions that really should be part of the core. I don't how much refactoring is enough to fix that. It would be nice to fix but I also don't want to do a half-baked refactoring that only makes things worse.

Hallbergs (approved):

PR #1639, opened 2025-04-24
Feature/1613 documenthandler open plugin map link

### LINKED ISSUE #1613: DocumentHandler: Open corresponding plugin on map-link click
A user has requested additional `map-link` functionality in `DocumentHandler`. Specifically in `MapviewModel`, the user should be able to add a link with`timeslider` activated to a document chapter and have the the timeslider (or any other component) open upon clicking the link:

Image

The result should be the same window opened as in the url link provided:

Image

To fix this addional logic for opening the plugins specified in the url (...p=documentviewer%2Ctimeslider) should be added to `MapviewModel`. 

The feature should only be available if `enableAppStateInHash` is true.

OlofSvahnVbg:
closes #1613


Hallbergs (changes requested):
Just a thought: I see that we only show plugins if we have `enableAppStateInHash` set. Couldn't/shouldn't we allow to show plugins in "regular" links as well? Even when `enableAppStateInHash` is disabled. 

Also, please update the changelog!

OlofSvahnVbg:
Seems to work by just removing the check for appState. However, I found something else. The flyTo seems to be a bit janky:

https://github.com/user-attachments/assets/1e36a46a-0126-4c03-9623-cded54440ea4

OlofSvahnVbg:
Added new function that updates url while enableAppStateInHash is enabled. Previous function now only runs when appState is disabled.

Hallbergs (approved):

OlofSvahnVbg:
So I don't think we need the updateUrl or a check if appStateHash is activated. This is because when the map link is set the url contains all the neccesary information **if appStateHash is activated** at the time of creating the map link. Therefore, maybe the check for appStateHash when clicking the map link is unnecessary.

Issue #1646, opened 2025-04-30
LayerSwitcher - Draw order


jesade-vbg:
When dragging and dropping in the layer order tab its flickering allot but it also doesn't sort as expected.
The problem escpecially occurs when System layers and background layers are visible.

The "effect" is clearly visible in the attached video.

https://github.com/user-attachments/assets/11021007-afe6-48e0-8e1c-530fcabd0c79

Hallbergs:
My guess is that several of the layers have the same z-index. We have check all system layers and make sure to set z-index on all of them, and also build some kind of handler for the layers from config...

jacobwod:
I agree with @Hallbergs. Either same z-index or a pre-coded _"hovered item's z-index +1"_ when they overlap, which may lead to the dropped item "jumping".

jacobwod:
Thanks for looking into that @michaelsund. You're right about `react-smooth-dnd` (which already seemed like an odd choice when it as added quite recently, given that the latest updates are 5+ years old). 

Feel free to look into alternatives! 👍 

Related: I just modified the build script and Dockerfile in apps/client, as it would not build (react-smooth-dnd wants React 16, not 16+, and we're on 19 currently…).

PR #1689, opened 2025-09-16
Replacement of react-beautiful-dnd with dnd-kit

michaelsund:
Since react-beutiful-dnd needs replacement I tinkered with first hello-pangea/dnd but soon realized it's limitations and unresolved bugs.
Dnd-kit is now being used to render the draworder list with drag and drop.
The resorting of the real layers in the map is still as is, but i cannot get the bug described in #1646 to trigger anymore.

Since i'm not very familiar with the draw-order in and outs, I could use some help testing it with all kinds of layers other than the default ones.


michaelsund:
Nice! Discovered another problem with the grouplayers not being able to expand the layers in draworder, I have a fix for that though, will do a separate PR later.

jacobwod:
No worries. I think you can do at least one more commit: looks like we're missing a package in package.json. 😉 

```
ERROR in ./src/plugins/LayerSwitcher/components/DrawOrder.js 5:0-60
Module not found: Error: Can't resolve '@dnd-kit/modifiers' in '/Volumes/CaseSensitive/hajk-v4/apps/client/src/plugins/LayerSwitcher/components'
```

jacobwod:
Feel free to press the merge button! 👍

Issue #1688, opened 2025-09-16
Quick Access Presets can't be disabled


jacobwod:
Even when you set what I believe is the correct setting in Admin, save and verify that it got saved in map config's JSON, you can't get the UI to hide the Quick Access part. See attached screenshot. (The other key that could have something to do with it, `enableUserQuickAccessFavorites` is also `false`.)

<img width="474" height="767" alt="Image" src="https://github.com/user-attachments/assets/4c65a5de-8483-412d-8269-e14de2665552" />

PR #1693, opened 2025-09-17
Fix for QuickAccess to not render when config says it should't.

michaelsund:
If I understood #1688 correctly, this should fix the admin option to not show the QuickAccess and pass the "show" prop along and conditionally render the QuickAccessView component or not.

Maybe it's better to conditionally render the whole component in LayersTab.


jesade-vbg (approved):
Looks good! I've tested both scenarios and it works as expected. THanx!

Issue #1643, opened 2025-04-30
LayerSwitcher - Legacy Info button and view visible


jesade-vbg:
Legacy Info button and view are still visible on some layers (Probably group layers).
It should be like the button marked with green, sending to the new view showing the info.

![Image](https://github.com/user-attachments/assets/e76935c4-f896-4537-9e3d-70cfcaf81db3)

![Image](https://github.com/user-attachments/assets/aec8bd7e-49a3-4be6-a66b-758eb4f342e2)

michaelsund:
Looks like the legend-groups created manually in the admin-ui are still handled in the LayerGroup component.

Would it be a good idea to modify the new view "BtnShowDetails" component to handle this info too but without the opacity/legend/add-to-quickaccess options?
And just show the info only?

Since it's not a real layer or grouplayer, setting the above mentioned stuff might be a bit tricky when you can also make these settings on the real group/layer.

PR #1699, opened 2025-09-18
Fix/layerswitcher legacy info

michaelsund:
Fix for #1643 to use a new info component similar to LayerItemInfo.

Discovered that "infogroupowner" is not present in the config even if set on a group in the admin-ui, but should probably be it's own issue.

Screenshot from 2025-09-18 21-00-35


jacobwod (changes requested):
I couldn't really test the functionality itself as my configs seem to lack this option. But here comes the code review.

I removed a couple of unsed imports in your code. Make sure to pull them.

In addition, I noticed this:
```
// LayerGroup.js, line 115
  const [infoVisible, setInfoVisible] = useState(false);
```

This is an older line that's been there before your fix. But I think you may have a better understanding of the compontent right now so perhaps you can decide if it should stay as a `useState`, or just be replaced with a basic `const infoVisible = false`. I can't see its value change anywhere in the code, but I may not have the full picture. Feel free to decide.

jacobwod (approved):
Works like a charm!

PR #1720, opened 2025-10-08
feat: Include LayerGroup names in LayerSwitcher search

karljakoblarsson:
There was a bug in the LayerSwitcher search where search hits on "`LayerGroup`" names did not show the group. (A `LayerGroup` is a group configured in the config. Not a layer with sublayers, which is called `GroupLayer`)

https://github.com/user-attachments/assets/5a088e4c-2c58-4a0d-a2fe-4264093eadf0


karljakoblarsson:
I'm not to happy with the variable names. I'll updated them if I can think of something better.

karljakoblarsson:
> Super! But please PR towards `develop` instead. 👍

Of course!

PR #1749, opened 2025-12-19
Fix/1748 bookmarks sublayer state

### LINKED ISSUE #1748: Saving a sublayer activates all sublayers in a group layer using Bookmarks plugin
**Describe the bug**  
When a bookmark is saved with only a specific sublayer of a group layer active, loading that bookmark activates all sublayers within that group layer in the `LayerSwitcher`, rather than only the originally selected sublayer.

**Steps to reproduce**  

1. Activate a sublayer in the `LayerSwitcher`.
2. Save the current map state as a bookmark.
3. Reload the application.
4. Load the previously saved bookmark.
5. Observe that the entire group layer, including all its sublayers, is activated in the `LayerSwitcher`, not just the sublayer that was active when the bookmark was saved.

**Expected behavior**  
When a bookmark containing an active sublayer of a group layer is loaded, only that specific sublayer should be activated.

linusfj:
Closes #1748.


Hallbergs (approved):
(y)

PR #1756, opened 2026-01-28
Feature/1731 user friendly cql filter

vagvaf:
Hi, here's the updated CQL filter. 

There is an attempt to get the field names from the WFS server so that users can choose the field name using a dropdown. When this is not possible, an input field is provided.

I have kept the original CQL filter input field so that someone can edit the filter directly  as well as see what the actual filter that is applied. Users have to click on the same button as before to apply the changes.


vagvaf:
> Please merge latest `develop` and perhaps remove the log message (line 25 in CQLFilter.js), thanks.

Hey, thanks. I spotted a bug. When a layer is not visible (unchecked) and one clicks on the dropdown to get its field names, the field names of all layers appear instead of only the field names belonging to that layer. I will make a fix and then merge the latest develop and also remove the log message.

PR #1796, opened 2026-04-13
Feature/1795 wmts implementation refresh

### LINKED ISSUE #910: Set zoom level for search service
When a search is made and you are zoomed to the object the zoom level is quite zoomed out for us which isn´t always desirable.

**Describe the solution you'd like**
The ability to set zoom level for search services in a map for each service in order to adapt it based on object that is searched for and the intended use of the map.

**Describe alternatives you've considered**
Set same zoom level for search results for a map so that all search services use the same level in that map.

### LINKED ISSUE #1795: WMTS - Implement it better (so it works).
WMTS does not work fully and allot of parameters are missing. Not fully implemented and forgotten.

This will be an attempt to get it working correctly and make Admin prefill data from GetCapabilities.
Then the client must ofcourse actually consume this data. 

Will be many changes and might not be backward compatible (Is anyone got it working before).

Image

jesade-vbg:
Closes #1795
Closes #910


jesade-vbg:
@eri-blan: Please report testing results here.

eri-blan:
Hi, We´ve tested this in our environment (Geodirekt with QGIS Server) and it seems to work there.

Hallbergs (approved):

PR #1797, opened 2026-04-13
Fixes visibility toggle on drag-n-dropped `kml` and `gpx` files

Hallbergs:
When drag and dropping files we we're not able to toggle the visibility of the imported files. 

Added a fix so that both `kml` and `gpx` files can be toggled.

Screenshot 2026-04-13 at 09 32 09


jesade-vbg (approved):
Works great, good work!

PR #1806, opened 2026-04-16
Feature/1804 infoclick link check 

### LINKED ISSUE #1804: Infoclick: hide links that point to non-existing resources
### Problem

Infoclick definitions often contain links built from feature properties (e.g. document URLs). These URLs aren't guaranteed to exist for every feature, resulting in broken links shown to the user. The only workaround today is adding a boolean flag in the source data for every link and wrapping it in `if`
### Solution

A `link-check` attribute on markdown links:

```
[Open document](document_url){link-check=true}
```

Links with this attribute start hidden. A HEAD request (5s timeout) checks each URL after render. If it returns 2xx, the link is shown. Otherwise it stays hidden. The HTTP status is stored in a `link-check-result` attribute for debugging.

Optionally, a `link-check-fallback` attribute can provide text to display in place of a broken link:

```
[Open document](document_url){link-check=true link-check-fallback="Document not available"}
```

Example rendered output after links are checked:

```html
<a href="https://brokenlink" link-check link-check-result="404">Im broken</a>
<span class="link-check-fallback">Document not available</span>

<a href="https://workinglink" link-check-result="200">Im Ok!</a>
```

Note: the working link has `link-check` removed (making it visible), while the broken link retains it (keeping it hidden via CSS).

jesade-vbg:
Closes #1804


jesade-vbg:
One way to test this can be seen here.
https://github.com/hajkmap/Hajk/issues/1804#issuecomment-4281074047

jacobwod (approved):
Here are the findings of my tests.

Prior your fix, a link is visible:

Image

But document is missing. Boohoo. :'( 

Image

---

With this fix, using the `fallback` option:

Image

While working links are there (I have tested both without and with fallback, hence the dupe here):

Image

And me is :') now. 

Go merge.

PR #1807, opened 2026-04-21
resetting window height on reopen, when minimized

### LINKED ISSUE #1802: Problem to expanding the tools after minimizing the tool
When you have activated a tool and then minimized the tool, only a small part of the minimized view is visible after another tool is opened over the same position. In other words, the minimized view becomes so small that it becomes very difficult to expand the tool again.

The error occurs on several existing tools in the current version but also on the new Attribute Editor.
It would be good if you have some time and troubleshoot what the error may be due to (assuming you have the same error in your environment).

What we have discovered is that the error occurs for the following tools:
- anchor
- sketch
- measure
- print
- bookmarks
- fmeserver
- location

For the following tools we do not get the same error:
- layerswitcher
- coordinates

michaelsund:
An attempt at solving https://github.com/hajkmap/Hajk/issues/1802.

Treating the window as closed on close, even if it's minimized.
Height is restored on window reopen.

### DIFF
--- a/apps/client/src/components/Window.jsx
+++ b/apps/client/src/components/Window.jsx
@@ -229,8 +229,17 @@ class Window extends React.PureComponent {
   }
 
   #onOpenChange(open) {
+    // On open, disregard previous states like minimized and reset the window height
     if (open) {
-      this.bringToFront();
+      if (this.state.mode === "minimized") {
+        this.setState({
+          height:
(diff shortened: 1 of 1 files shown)


jesade-vbg:
Looks good! thanx!

jesade-vbg (approved):
Thanx! Great!

PR #1811, opened 2026-04-24
Print: output PDF with legends (experimental feature)

### LINKED ISSUE #1810: Print: Experimental legend support in PDF output
It would be useful to include map legends when generating PDF prints.

This is proposed as an experimental feature, since legend rendering can vary significantly depending on server implementations (e.g. WMS capabilities, styling differences, etc.). Because of this variability, the feature should not be considered production-ready initially, but rather something that can evolve and mature over time.

We need to start somewhere right?

Let me know if my reasoning seems off.

Here is an example of the initial test (Just activated random layers...)

Image

Image

Image

Image

Image

Image

jesade-vbg:
Closes #1810

Keep in mind that this is an experimental feature.

Focus when testing should (in this state) be so that we make sure the previous functionality stays the same and nothing old has been broken because of these code changes.

### DIFF
--- a/apps/admin/src/views/tools/print.jsx
+++ b/apps/admin/src/views/tools/print.jsx
@@ -51,6 +51,8 @@ var defaultState = {
   mapTextColor: "#000000",
   useCustomTileLoaders: true,
   includeImageBorder: false,
+  allowLegendsInPdfOutput: false,
+  generateLegendsByDefault: false,
   maxTileSize: 4096,
 };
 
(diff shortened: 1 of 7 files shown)


jacobwod:
I have now tested with this option _disabled_. I cannot see that it affects the existing printing in any way.

Do you want me to test the other parts too by enabling it?

jacobwod (approved):
Does what it says. Limited to PDF, but you write it out loud and make the option disappear when switching to PNG, so that's very correct too. Feel free to merge!

Image

jacobwod:
If I'm about to mention something that isn't optimal it would be that **column break would be nice, so we use the entire paper width for legends**. But I understand it's difficult to achieve so nevermind.

image

PR #1813, opened 2026-04-27
Feature/1812 layer comparer spy window resize

### LINKED ISSUE #1812: LayerComparer: Change size of Spy Window etc.
The spy thing in LayerComparer has already got additional requests.

We need to be able to resize the spy thing.

jesade-vbg:
Closes #1812


jacobwod (approved):
Insanely cool!

PR #1817, opened 2026-05-01
Feature/toggle label layer

### LINKED ISSUE #1816: Allow dynamic labels toggling in map for active layers
We're getting requests to mimic the functionality available in QGIS that allows users to toggle labels in map on and off. 

Image

Image

One way this could be solved is by having multiple styles (in the WMS service) - one with and one without labels.

While doable, it's cumbersome now, because a layer's style is hard-coded in the layer definition in Hajk admin. While we allow selecting a style (and it gets sent to the WMS service as `STYLE=xyx`), it's still one-per-Hajk-layer. We'd need to open up the UI somehow, as well as listen to that value and pass it on to the WMS service.

michaelsund:
Separate PR for the client adding the useLabelLayer key to the layersConfig > wmslayers array. (backend/admin part handled separately)
For the label button to be show  we now expect the useLabelLayer key to exist and have the boolean value of true, otherwise its disregarded. So this should not affect existing configurations that's missing this key.

As discussed [here](https://github.com/hajkmap/Hajk/issues/1816) we expect that the wms source has a STYLE available named "labels". If the STYLE is added with any other values and is not left blank, we save these and reapply them when the label layer should no longer be visible.

`{
        "id": "1353",
        "caption": "Polygons of interest",
        "url": "http://192.168.1.2:8086/qgis-server",
        ...
        "visibleAtStart": true,
        "hasLabelLayer": true,
        etc..
`

The label can be activate before the layer is activated, showing the label layer directly.
Although when the layer is disabled, the label layer is disabled aswell to "reset" the menu entry to it's default state.

https://github.com/user-attachments/assets/3dc2461c-72b1-40ed-b313-fa0eeb510b09

### DIFF
--- a/apps/client/src/models/layers/LayerInfo.js
+++ b/apps/client/src/models/layers/LayerInfo.js
@@ -30,5 +30,6 @@ export default class LayerInfo {
     this.layerType = properties.layerType;
     this.hideExpandArrow = properties.hideExpandArrow;
     this.showAttributeTableButton = properties.showAttributeTableButton;
+    this.hasLabelLayer = properties.hasLabelLayer;
   }
 }
--- a/apps/client/src/models/layers/WMSLayer.jsx
+++ b/apps/client/src/models/layers/WMSLayer.jsx
@@ -18,10 +18,6 @@ class WMSLayer {
     this.attribution = config.attribution;
     this.layerInfo = new LayerInfo(config);
     this.subLayers = config.params["LAYERS"].split(",");
-    this.globalObserver.subscribe(
-      "layer.toggleLabelLayer",
-      this.onToggleLabelLayer
-    );
 
     const source = {
       url: config.url,
@@ -131,20 +127,6 @@ class WMSLayer {
     this.bindHandlers();
   }
 
-  onToggleLabelLayer = ({ layerId }) => {
-    if (layerId !== this.layer.get("name")) return;
-    console.log(layerId);
-    const source = this.layer.getSource();
-    const params = source.getParams();
-
-    const currentStyles = params.STYLES || "";
-    // The naming convention for our wms label layers are "labels"
-    const isLabelActive = currentStyles === "no_labels";
-
-    source.updateParams({
-      STYLES: isLabelActive ? "" : "no_labels",
-    });
-  };
   applyCustomDpiTileLoader(source, config) {
     // Experimental
     //
--- a/apps/client/src/models/layers/WMSLayer.jsx
+++ b/apps/client/src/models/layers/WMSLayer.jsx
@@ -122,6 +122,7 @@ class WMSLayer {
         : this.subLayers
     );
     this.layer.set("hasLabelLayers", config.hasLabelLayers);
+    this.layer.set("showLabelLayer", false);
     this.layer.getSource().set("url", config.url);
     this.type = "wms";
     this.bindHandlers();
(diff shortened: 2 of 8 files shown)


jacobwod:
Looks very promising @michaelsund! One thought though: as I understand it, the current requirement is that the WMS layer must have an associated style named exactly `labels` for this to work? If so, I imagine an issue here, where multiple layers in the same WMS workspace would have to have the same style, albeit we don't really want it to be the same style. E.g. layers `fastigheter` and `byggnader` would both need to be styled with the style `labels`. 

Don't you think the style name requirement should rather be something like `<layerName>_labels`?

michaelsund:
@jacobwod I think i kinda forgot our reasoning when i setup my qgis-server layer and looked at the ones you linked who used styles to differentiate the layers :smile:

I guess i got ahead of myself and zoned in on using styles, now in hindsight i guess having the suffix `_labels` would duplicate the amount of layers though in the wms (that should use labels), whereas using styles would keep the wms layerlisting cleaner?

It feels like either handling multiple copies of the same layer for each layer in a workspace, or handling multiple styles for each layer. Which one do you think we should choose?

jacobwod:
Sorry if I was being unclear: no, I don't think we should duplicate layers in the WMS. The only thing that should be needed is 1) a layer and 2) two styles associated with it (one being the default style, used when none is provided in the `STYLE` parameter to GetImage, the other being the optional style with labels).

michaelsund:
Ah! pushed an update that does just that.

jacobwod (dismissed):
Works perfect. If I'd think of something to add, it'd be a minor thing, like a tooltip over the button in LayerSwitcher that explains what this button does. Otherwise, it's perfect!

https://github.com/user-attachments/assets/beb27e15-5cfe-4d15-bafe-2735fe05e6f1

michaelsund:
Good idea! I added it below though as the HajkToolTip component doesnt seem to take a position prop.
The other buttons in the layerlist have the tooltip appear below, so i went with that.

jacobwod:
Super! Now on further testing I can't seem to activate this using the Draw order tab (not sure if you have it active, but if you have you see the exactly same row and button next to it). I haven't had time to investigate further but it seems like nothing happens when I click that button - is there some event listener not being registered on this tab?

image

michaelsund:
Hmm, i didn't think it would appear there so I honestly didn't check the draworder.
I guess it's because they both use the same LayerItem component but pass props diffrently.

I'll investigate.

@jacobwod Looks like they component is used there too, I can wire up the props so it works in DrawOrder too, but they will have separate internal states which is not optimal.

Do we want to show this button in the DrawOrder too? In that case I need to lift the state up or store it in the olLayer settings.

Edit: I'm thinking yes since we show the "More" button in the DrawOrder aswell, they should look the same in both LayerSwitcher and DrawOrder, and with the recent push the label layer can be toggled from both places.

jacobwod (approved):
This works excellent and I am going to approve it. 

The one limitation, that in my opinion is out-of-scope for this initial release, but that I can foresee users asking for in the future is _the ability to load map with labels on/off_. As you know, we have the query string parameters that control how map loads on start (e.g. the `l` parameter tells Hajk which layer [ids] to load). Like this: https://karta.halmstad.se/#m=ext&x=110593.95010035105&y=6283798.881446072&z=10&l=uk1pqo%2Ct7z9k6&p=layerswitcher

In order to get this working, I think that we would need to introduce another query parameter, or some clever notation within the existing `l`-parameter. Something like `l=uk1pqo` would load the default map, but if we assume that layer ids never include underscores, we could have `l=uk1pqo_l` for loading layer `uk1pqo` with the labels active. 

Now, this was a long review for a PR I'm approving in its current state anyway. 😄

PR #1819, opened 2026-05-04
Add support for CQL filtering in admin interface

### LINKED ISSUE #1818: CQL filtering support in admin
Currently, `CQL filtering` is supported on the client side, but it is not possible to configure or apply `CQL filtering` through the admin interface. This creates a gap between what regular users understand and what administrators are able to configure.

As a result, admins cannot define dynamic filters for WMS layers (e.g., from `Geoserver`) without relying on client implementations or creating multiple WMS services for different filtering needs. This often leads to unnecessary duplication of layers or datasets in `Geoserver`and `PostGIS`, increasing maintenance complexity.

**Describe the solution you'd like**
Add support for configuring `CQL filters` directly in the admin interface, aligned with the existing client functionality. This could be implemented as:

An input field for defining CQL expressions

**Enabling CQL filtering in admin would provide several benefits:**

Reduces the need for duplicate WMS layers or datasets
Minimizes redundancy in `Geoserver` and `PostGIS`
Leverages the full potential of the WMS standard, where `CQL filtering` is an established and intended feature

_EDIT_
_**This also works if the CQL filter in `LayerSwitcher` is off**_

**Preview**

Image

** automatically updates cql in client**

Image

Image

Albinahmetaj:
closes #1818


jesade-vbg (approved):
Good addition! Good work!

PR #1821, opened 2026-05-05
Feature/documenthandler iframe support

### LINKED ISSUE #1820: Documenthandler: Add iframe support for dynamic content embedding (e.g. Microsoft Power BI)
Currently, the `documenthandler` does not support embedding dynamic external content, meaning visualizations such as charts or dashboards need to be updated manually or managed outside the system.

Add support for embedding external content via iframe in the `documenthandler`. This allows admins to configure and include dynamic content (e.g., Microsoft Power BI reports) directly within text documents.

This enables:

-Integration of live, data-driven visualizations within documents
-Automatic updates when the underlying data changes
-Support for tools like Power BI to present dashboards and analytics

One example use case is ÖP follow-up, where the current status of projects can be visualized (e.g., number of ongoing, completed, and pending projects). Since these values change over time, this highlights the need for live, continuously updated visualizations rather than static representations.

Preview (admin):

Image

I have also added a preview of the iframe in edit mode

Image

**Client**

Image

Albinahmetaj:
Closes #1820 

We have been trying this inhouse and it seems to work fine. We have not identified any performance issues during testing
However, I’d appreciate it if more people could test it.


jacobwod (dismissed):
Seems to work fine. We should keep in mind that `X-Frame-Options` and CSP can cause problems when embedding from different origin than where Hajk is deployed, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Frame-Options and https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/frame-ancestors. 

Remember that this is controlled by the embedded site (i.e. _not_ by our Hajk instance, but the resource we want to embedd), so having this work requires the ability to control HTTP response headers on that server. Something like this is what is needed:

```HTTP
Content-Security-Policy: frame-ancestors 'self' https://server.that.runs.hajk.org
```

jacobwod:
BTW @Albinahmetaj , don't forget to add a line to CHANGELOG.md

Albinahmetaj:
Will do, thank you for the review

jacobwod (approved):

PR #1823, opened 2026-05-07
Load layers with labels on app start.

michaelsund:
As @jacobwod described [here](https://github.com/hajkmap/Hajk/pull/1817) users will probably want to load the map with the labels toggled on/off to reflect the same view one user may link to another user..

We reserve the suffix of the layername in the querystring, where `{layername}_l` indicates that the layer should have it's _labels style loaded.

A caveat is that if a layer is linked with the _l suffix in the url's querystring under key "l", for example mylayer_l.
And mylayer doesn't support labels, the whole layer is removed from the querystring on app start and the "normal" layer is not activated at all.

If this should be the default behavior should be decided, or if the normal layer should be activated as a fallback.

Since there's some changes done in the AppModel, I would appreciate help testing this PR since the loading of WMS layers have been modified and could affect plugins/modules I'm not familiar with.

### DIFF
--- a/apps/client/src/models/AnchorModel.js
+++ b/apps/client/src/models/AnchorModel.js
@@ -106,7 +106,15 @@ class AnchorModel {
           isValidLayerId(layer.getProperties().name)
         );
       })
-      .map((layer) => layer.getProperties().name)
+      .map((layer) => {
+        // Add a check if we want this layer to use labels or not
+        // If so, append _l to the layer in the url
+        const name = layer.getProperties().name;
+        if (layer.get("showLabelLayer")) {
+          return `${name}_l`;
+        }
+        return name;
+      })
       .join(",");
   }
 
--- a/apps/client/src/models/AnchorModel.js
+++ b/apps/client/src/models/AnchorModel.js
@@ -53,10 +53,18 @@ class AnchorModel {
         const layerId = layer.get("name");
 
         // Update anchor each time layer visibility changes (to reflect current visible layers)
-        layer.on("change:visible", async (event) => {
+        layer.on("change:visible", async (_event) => {
           this.#app.globalObserver.publish("core.mapUpdated", {
             url: await this.getAnchor(),
             source: "layerVisibility",
+          });
+        });
+
+        // E: Update anchor when label layer state changes
+        layer.on("change:showLabelLayer", async (_event) => {
+          this.#app.globalObserver.publish("core.mapUpdated", {
+            url: await this.getAnchor(),
+            source: "labelLayerToggle",
           });
         });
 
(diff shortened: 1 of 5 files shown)


jacobwod:
Hi @michaelsund, really nice to see you are about to fix this as well!

I thought about this yesterday and came up with a slightly different solution, involving _a new key in the query string, e.g. `ll` ("layers with labels" 😉), that would basically be a comma-separated string of layer ids_. Yes, this means that to activate a layer with labels on, it would have to be present in both the `l` and `ll` parameter. So that's not nice perhaps. But it wouldn't be an issue if non-existing layers or misconfigured layers (without the label style) would be there: you can easily have something like:
```
l=layerWithoutLabels,layerWithLabels,missingLayer&ll=layerWithLabels,missingLayer,fooBarFakeLayer
```

And this would all work out well by checking with what's in `mapConfig` and end up with:
```
l=layerWithoutLabels,layerWithLabels&ll=layerWithLabels
```

Meaning that map starts with `layerWithoutLabels` and `layerWithLabels` visible at start, and that the latter has its labels style activate.

---

Your approach is good too and perhaps could work in a similar way: 

```
l=layerWithoutLabels,layerWithLabels_l,missingLayer
```

Gets transformed and the map loads as in my previous example, with two layers active where the last one has a label style active.

I don't see an issue where we accidentally have `_l` as a suffix in layer ids - that's just not how they are generated (it's only a-z & 0-9). 

So your approach could work very well, except I think we still should show a layer with the `_l` suffix, even if the labels style is missing. Just remove the `_l` suffix and if the "regular" layer ID exists, set it to visible.

What do you think?

michaelsund:
Oh! :smile: Well they sound similar enough then, except the extra queryparams.

I agree, that's how it should work and I discovered by testing a bit now that it does! If you paste the url into a new tab (ie load the app from scratch) it gets rewritten to a normal layer and is visible if useLabelLayer is false.

I'm having various problem occur when the url is updated and enter is pressed, a useLabelLayer enabled layer not getting updated whet _l is added, and non-enabled layers with _l added gets removed completely.
Probably something that isn't handled or updated after the initial app load.

I doubt someone would change the layers in the url in a normal use case, but there could be a scenario where a user tries to paste a hajk-link into an existing tab where hajk is loaded, which would yield the same problems.

I'll try to find a solution for this.

jacobwod:
Ah I almost forgot about it but you're right: currently user (or perhaps more probable: a webpage that embeds Hajk and controls it using the IFRAME's `src` parameter) can change a value in the query string, e.g. modify the `p` parameter to control which plugin is visible, or change the `l` to control layers' visibility. I just tested (those two cases, to be specific) and it works. So you're right regarding that we probably want a similar change (e.g. the `l` paramter changes from `abc123_l` to `abc123`) to actually trigger the labels style switch.

michaelsund:
- Layers can now be controlled again from the url, with or without _l, and fallback
    for layers that should not be able to show _l.
  - The above applies to pasting a layer into a empty window/tab or manually updating
    the url and hitting enter.

  When manually editing l= for a layer that doesnt support labels hasLabelLayer false
    ex l=mylayer and _l is added and enter is pressed, the url isn't rewritten immediately
    until a layerchange event is emitten (which i cannot pinpoint atm), the feature still
    works as expected though.

So far I have only found Bookmarks to not work with label layers, I'll look into that next.

jacobwod (approved):
This looks and works really well @michaelsund . I merged some changes from `develop` and renamed one of the flags used. Hope you're ok with it. Ready for merge imo.

jacobwod:
I think we should merge this real quick as the version that's currently in `develop` has a bug resulting in any custom style set in Admin to be ignored in Client… 😄  Luckily fixed in this branch already.

Can you merge @michaelsund or is it that you don't have the permission?

michaelsund:
Oh! Yes let me hit that button then 😃

PR #1824, opened 2026-05-08
TypeScript in Client: PropertyChecker overhaul

### LINKED ISSUE #1761: Property Checker: Requirements for public release
As the Property Checker is about to be released to the general public, there are several requirements that must be fulfilled:

* **Configurable reporting:** We need the ability to turn the reporting section off completely via settings. 
* **Layer descriptions:** Each "hit layer" should include a short description explaining what it is. Currently, only the name is shown. Note that these descriptions should be independent of the ones used in the Layer Switcher.
* **User guidance:** The tool needs support for additional descriptive texts. We need to be able to inform the user about the tool's purpose, provide instructions on how to use it, and help them understand how to interpret the results.

jacobwod:
I create this PR not so much because of the PropertyChecker updates (guess nobody except us uses it yet anyway), but because it marks the first use of TS in Hajk's Client app. 

I think the ground should be pretty solid and I verified it by converting the PropertyChecker plugin to TS. Of course some system-wide changes were necessary:
- TS fixes for ESlint
- Changes to how we load modules in AppModel (it's not just .js now, we must look for .ts too)
- Type definition for common files. Most of it goes to `apps/client/src/types`, e.g. `apps/client/src/types/hajk.d.ts`. Other should be placed "close to its usage", e.g. `BaseLayerWindow.d.ts` or definitions for specific plugins. 

Feel free to post your opinions. I await some other PRs before merging as I don't want to cause merge problems for anyone given changes in `AppModels.js`.

Anyway, closes #1761.


michaelsund:
Not using property checker, but the most common plugins and the build in tools seems to work good!

Hallbergs (approved):
I like your stance on the type declarations!

Albinahmetaj (approved):
Looks like a solid first step bringing TS into the client app. I also like the approach of keeping interfaces in separate type definition files. Good work!

PR #1825, opened 2026-05-12
Fixes bookmarks with layers that should show labels.

michaelsund:
An attempt to fix bookmarks showing layers with labels, a bit hacky to use a setTimeout for 100ms in . But we need to wait a tiny bit for LayerItem to mount and be ready before we set the wms style.

The usage of useCallback hook in LayerItem and the accompanied eslint rule disable is needed since useCallback in a useMemo component means that the react compiler can't do its thing to optimize, and eslint frowns.
Anyway, it stops applyLabelStyle from not rerunning during re-renders where it shouldn't.

### DIFF
--- a/apps/client/src/models/AppModel.js
+++ b/apps/client/src/models/AppModel.js
@@ -711,8 +711,10 @@ class AppModel {
 
         if (layer.hasLabelStyle === true) {
           layerItem.layer.set("hasLabelStyle", true);
+          // Store the layername so we can access it before wms layer changes
+          // ex when switching labels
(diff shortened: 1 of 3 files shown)


jesade-vbg:
Can you test if requestAnimationFrame would work instead of setTimeout? Its a little less hacky than a timeout guess.

michaelsund:
RequestAnimationFrame did now work in my tests, but storing the changed layername we are switching to in the layer makes it so that LayerItem's applyLabelStyle can update the wms params correctly in time.

jacobwod (approved):
I tested here, it's ready to merge IMO.

PR #1828, opened 2026-05-20
Merge/develop into admin-ui/main

jacobwod:
server.ts has a lot of new stuff in this branch, related to Prisma, express-session etc, that we didn't have in develop (and still don't). However, some changes from develop have now been incorporated to the server.ts in this branch, so we're kind of in sync. 

Client, and the legacy Admin, fast-forwarded without conflicts.


Albinahmetaj (approved):
LGTM

PR #1829, opened 2026-05-21
Client: Fix all ESlint problems

jacobwod:
Here's an attempt to go from 400+ ESlint problems to ~zero~ ...less than that. 

~This is still a draft and I'll commit as I proceed.~

There are some 120 errors left but I think we should take it in smaller steps from here on, hence I propose to merge this now.

### DIFF
--- a/apps/client/src/components/Alert.jsx
+++ b/apps/client/src/components/Alert.jsx
@@ -14,7 +14,7 @@ class AlertView extends React.PureComponent {
     parent: propTypes.object.isRequired,
   };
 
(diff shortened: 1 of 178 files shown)


jesade-vbg:
I have slowly scrolled thru the whole massive PR (even thou its a draft) and think it looks all good. Cant see anything that is even a warning flag.
Good work with this cleanup! 💯

jesade-vbg (approved):
Looks great! Works fine 💯

PR #1830, opened 2026-05-22
Enforce ESLint and Prettier in VSCode/Cursor

jacobwod:
Commits two small changes to `.vscode/` that make ESLint and Prettier violations show up as red underlines in VSCode and Cursor for everyone, not just those who happened to configure their editor correctly.

Here's what actually changed and why:

- .vscode/settings.json — tells the ESLint extension to use flat config, auto-fix on save, and use Prettier as the default formatter for non-JS/TS files. Since we already run Prettier as an ESLint rule (prettier/prettier: "error"), a single ESLint fix-on-save handles both linting and formatting for JS/TS files.
- .vscode/extensions.json — recommends the correct extensions which makes VSCode/Cursor prompt you to install them when opening the repo.
- .gitignore — un-commented the extensions.json exception so the new file is actually tracked.

Old Admin UI is intentionally excluded since it's being replaced.


jesade-vbg (approved):
Works like a charm! 🥇

PR #1835, opened 2026-05-27
Feature/1833 documenthandler maplink link search

### LINKED ISSUE #1833: DocumentHandler: maplink and link are not triggered when opening a document from search results
## Bug description

When a document menu item has `maplink` (map zoom) and/or `link` (external URL) configured, these connections are correctly triggered when the user clicks the item in the panel menu. However, when the same document is reached by clicking a search result, only the document is opened and the map does not zoom and no external link is opened.

## Steps to reproduce

1. Configure a document menu item with both a `document` and a `maplink` (and/or `link`) in the map config.
2. Open the map and use the search box to search for a heading inside that document.
3. Click the search result.
4. **Expected:** The document opens at the correct heading, the map zooms to the map link, and/or the external link opens in a new tab.
5. **Actual:** The document opens at the correct heading, but `maplink` and `link` are silently ignored.

## Root cause

The normal panel-menu click flow calls `#openEventualConnections(id)` in `PanelMenuContainerView.jsx`, which sequentially triggers all three handlers:
- `#handleOpenDocumentFromPanelMenu(id)` – opens the document
- `#handleExternalLinkClicked(id)` – opens external link if present
- `#handleShowMapLayersFromPanelMenu(id)` – zooms the map if present

The search click flow follows a different path:

1. `DocumentSearchModel.js` handles the `search.featureClicked.documentHandlerSearchResultClicked` event and publishes `"document-link-clicked"` with only `{ documentName, headerIdentifier }` — no `maplink` or `link`.
2. `PanelMenuContainerView.jsx` receives `"document-link-clicked"` and calls `#handleOpenDocumentFromLink()`, which only calls `#setDocument()` and scrolls to the menu item. The external link and map link handlers are never called.

## Suggested fix

In `#handleOpenDocumentFromLink()`, after finding the matching menu item (`itemClicked`), also trigger the external link and maplink handlers using the item's `id`, similar to how `#openEventualConnections(id)` works:

```js
#handleOpenDocumentFromLink = ({ documentName, headerIdentifier }) => {
  const itemClicked = findMenuItemWithDocumentName(documentName, this.state);
  this.#setDocument(documentName, headerIdentifier);
  this.#handleExternalLinkClicked(itemClicked.id);
  this.#handleShowMapLayersFromPanelMenu(itemClicked.id);
  this.#setItemStateProperties(itemClicked.id).then(() => {
    this.#scrollToMenuItem(itemClicked.itemRef.current.offsetTop);
  });
};

### PR DESCRIPTION
**Pull request by Sunix71**
## Summary

- maplink and link were not triggered when a document was opened via search results, only when clicked in the panel menu.
- Root cause: #handleOpenDocumentFromLink() only opened the document and it never called the external link or maplink handlers.
- Fix: two additional calls to #handleExternalLinkClicked() and #handleShowMapLayersFromPanelMenu() in #handleOpenDocumentFromLink() in PanelMenuContainerView.jsx.

## Test

- [ ] Configure a document menu item with maplink and/or link in the map config
- [ ] Search for a heading inside that document
- [ ] Click the search result and verify the document opens at the correct heading, the map zooms to the maplink, and/or the external link opens in a new tab
- [ ] Verify normal panel menu click still works as before

Fixes #1833

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Client: DocumentHandler - Fixed maplink and link not being triggered when opening a document via search results. [#1833](https://github.com/hajkmap/Hajk/issues/1833)
 - Client: DocumentHandler - Fixed subscription accumulation in Contents.jsx that caused duplicate print headers [#1773](https://github.com/hajkmap/Hajk/issues/1773)
 - Client: Ensure the Quick Access function respects the cookie setting [Issue #1798](https://github.com/hajkmap/Hajk/issues/1798)
 - Backend: Removed/replaced two unnecessary dependencies, see [commit](https://github.com/hajkmap/Hajk/commit/138e8668326b19a643542330ffef2ec5e3d847a6).
--- a/apps/client/src/plugins/DocumentHandler/panelMenu/PanelMenuContainerView.jsx
+++ b/apps/client/src/plugins/DocumentHandler/panelMenu/PanelMenuContainerView.jsx
@@ -83,10 +83,10 @@ class PanelMenuView extends React.PureComponent {
   #handleOpenDocumentFromLink = ({ documentName, headerIdentifier }) => {
     const itemClicked = findMenuItemWithDocumentName(documentName, this.state);
     this.#setDocument(documentName, headerIdentifier);
-    this.#handleExternalLinkClicked(itemClicked.id);
-    this.#handleShowMapLayersFromPanelMenu(itemClicked.id);
     this.#setItemStateProperties(itemClicked.id).then(() => {
       this.#scrollToMenuItem(itemClicked.itemRef.current.offsetTop);
+      this.#handleExternalLinkClicked(itemClicked.id);
+      this.#handleShowMapLayersFromPanelMenu(itemClicked.id);
     });
   };
 
--- a/apps/client/src/plugins/DocumentHandler/panelMenu/PanelMenuContainerView.jsx
+++ b/apps/client/src/plugins/DocumentHandler/panelMenu/PanelMenuContainerView.jsx
@@ -83,6 +83,8 @@ class PanelMenuView extends React.PureComponent {
   #handleOpenDocumentFromLink = ({ documentName, headerIdentifier }) => {
     const itemClicked = findMenuItemWithDocumentName(documentName, this.state);
     this.#setDocument(documentName, headerIdentifier);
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by jesade-vbg** [comment]
Good work, but I notice something weird when I click in the search the header jumps up a bit and gets miss-aligned. And it looks like that until you refresh the page.

image

**Comment by Sunix71** [comment]
Thanks for testing! 
Could you confirm if the document you tested has a maplink configured in the map config?
The header jump might be related to my fix, but only if maplink is set
I'd like to verify before investigating further.

**Comment by jesade-vbg** [comment]
Its confirmed, When I click the item in Search the header gets messed up. When I click the ordinary navigation it does not happen.


``` json
               {
                  "title": "Vad är en översiktsplan?",
                  "document": "1_1_om_op",
                  "color": "#edc3d2",
                  "icon": {
                    "materialUiIconName": "",
                    "fontSize": "large"
                  },
                  "maplink": "/#m=map_1&x=150000&y=6341002&z=0&l=1423%2C1450%2C1404%2C1405%2C1406%2C1407%2C1409%2C1410%2C1411%2C1448%2C1437%2C1412&p=documentviewer",
                  "link": "",
                  "menu": [],
                  "expandedSubMenu": false
                }
```


Notice that the drawn green arrows below are 100% man made, no AI has been used to produce this work of art.
image

Sunix71:
I was not able to reproduce the issue on my end, but I think I've found the cause, perhaps.
The two new calls were placed before the setItemStateProperties promise, meaning the maplink could trigger a layout change asynchronously before the menu scroll had a chance to anchor itself.
I've moved them inside the .then(), so the scroll runs first and the maplink fires after. 
A bit of blind coding on my part though, could you test and see if it helps?
Sorry for the inconvenience!

jesade-vbg:
No change Im afraid. Weird that only I get the problem. Maybe someone else should test.

Sunix71:
I've been able to reproduce the issue! On my end it seems to be related to "resizingEnabled": true in the DocumentHandler options in map_1. 
When I set it to false the header no longer jumps. 
I don't think this is caused by my change, but I'd love for you to test it and confirm, does setting resizingEnabled to false fix it for you as well?

Would you like me to open a separate issue for the resizingEnabled bug?

The same bug appears in the develop-branch

jesade-vbg:
Ok, good research! Confirmed!

Yes, please open a separate issue for the resizingEnabled bug.

jesade-vbg (approved):
Looks good! Works fine! 

Good Work!

jesade-vbg:
FYI: You have lost the "Verified" status on your first commits so I'll need to force merge this.

PR #1837, opened 2026-05-27
Upgrade MUI to v9

jacobwod:
Unusually complicated upgrade this time but at last it starts and runs without console errors, so ready to merge. I have not tested anything with the `@mui/x-data-grid` component however.


jacobwod:
Let's wait a second with this, I'm not so sure that the merge in https://github.com/hajkmap/Hajk/pull/1837/commits/ea4b1e3183d5b42f19c3794242f5ec052b4a94ba is right, it seems to add back some stuff as `sx` props that I have previously moved to one of the `slotProps` attributes.

apps/client/src/plugins/DocumentHandler/documentWindow/ImagePopupModal.jsx
jesade-vbg, line 1: This should probably be checked and verified an extra time

jesade-vbg: It seems to be working as exprected.

jesade-vbg (approved):
Tested and seems to be working just fine 💯 Good job
Cant notice an new visual bugs/changes or logic changes.
I've gave ImagePopupModal.jsx some extra checks as the new logic looked fragile. 👍🏻

jacobwod:
This PR is superseded by #1840 because I wasn't sure that the merge of `develop` yielded intended results here. So I went back to MUI 9 branch _prior_ merge of `develop` and rebased those changes on the lastest `develop`. The results were pushed to a new branch. Those two branches differ so meaning the two methods gave different results, which is interesting. See diff here. https://github.com/hajkmap/Hajk/compare/chore/upgrade-mui-v9...chore/upgrade-mui-v9-rebase-develop

Issue #1831, opened 2026-05-25
DocumentHandler's window shows color marker even when collapsed


jacobwod:
All windows in Hajk have a "highlight color" bar between the window's toolbar and main content area. See below:

<img width="1149" height="526" alt="Image" src="https://github.com/user-attachments/assets/2c6e3452-64b8-4764-a940-df39f83f1759" />

When a window gets collapsed, it should only show the toolbar (as LayerSwitcher window does in the example below). DocumentHandler's window does not follow this procedure and shows the color bar even when collapsed. This should be fixed.

<img width="1145" height="390" alt="Image" src="https://github.com/user-attachments/assets/846fc986-764d-426c-ad4b-5b1e8706c209" />

--- 

_Up for grabs, as I'm into unrelated code right now._

PR #1855, opened 2026-06-24
Fix for issue 1831 window collapse should hide color marker.

michaelsund:
As per [Issue 1831](https://github.com/hajkmap/Hajk/issues/1831)

This is an attempt to fix the hiding of the color marker on all windows that uses the PanelHeader. Albeit some windows had their color marker already not visible on collapse, examples like Bookmarks and the Document Handler window did show their color marker.

Screenshot from 2026-06-24 18-19-27


jesade-vbg (approved):
Looks good! Good job!

PR #1857, opened 2026-06-25
Fixed the print issue in #1856

### LINKED ISSUE #1856: Bug: Client Print: Ugly Title positioning and alignment etc.
Print titles and subtitles are positioned incorrectly: they are left-aligned (and wrapped lines aren't centered) instead of being horizontally centered.

The subtitle uses a fixed vertical position, so it overlaps or drifts away from the title when the title wraps to multiple lines. The title's height isn't calculated for wrapped text, so multi-line titles aren't handled cleanly.

Before:
Image

After potential fix:
Image

jesade-vbg:
Closes #1856

Not the prettiest solution but there not many alternatives.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


Albinahmetaj (approved):
I was able to reproduce the issue described in #1856 , and this PR resolves it. LGTM!

PR #1861, opened 2026-06-27
MapSwitcher dropdown on desktop

### LINKED ISSUE #1858: MapSwitcher: make it easier to see and change map config
Our users have identified a potential improvement when working with different map configurations in Hajk.

As you know, one instance of Hajk can have multiple map configs. User can switch between them using a dropdown that appears when clicking on a _control button_:

Image

Image

---
**Oftentimes, we get requests from users that finally end up with the answer "you were being using wrong map"**. 

To solve this, we propose that the MapSwitcher control button functionality is expanded. Admin should be able to decide if it should show as-is now (small control button) or as a larger dropdown element, perhaps next to the Search bar (or elsewhere, exact placement to be determined):

Image

In that case, the small control button does not show, as in mockup above.

Also, this is desktop-only. Small screens should probably be unaffected.

_This is an idea - let me know your opinions!_

michaelsund:
Suggested fix for [1858](https://github.com/hajkmap/Hajk/issues/1858).

The dropdown in configurable in the admin tool per map.
"Visa som knapp" or "Visa som rullista i sidhuvud".
Screenshot from 2026-06-27 11-43-07

The dropdown is hidden on mobile and the button is shown instead in it's normal place (even if the map is set to "Visa som rullista i sidhuvud").
This is currently not responsive since isMobile is evaluated once when App.jsx is mounted.

Screenshot from 2026-06-27 11-57-39


jacobwod:
This works perfectly!

Regarding having some kind of label over the dropdown, we can have something like this:
Skärmavbild 2026-06-29 kl  10 55 24

I'm not sure if it clutters the UI a lot or not. Perhaps it's tolerable and gives new/non-frequent users a grasp of what this dropdown does?

I'll commit this change and let you decide @michaelsund.

michaelsund:
That InputLabel is perfect, small enough to still convey what the dropdown is for.

PR #1864, opened 2026-06-29
Version 1.0 of LayerControlModel

### LINKED ISSUE #1863: Core: Centralized Layer Toggle API
## Centralized Layer Toggle API

Create a `LayerControlModel` for toggling layers by id.

It will support normal layers, WMS group layers/sublayers, base layers, and LayerSwitcher group folders.

Expose it as `appModel.layerControl` and `window.hajkPublicApi.layerControl`.

Required methods: `showLayer`, `hideLayer`, `toggleLayer`, `layerIsVisible`. (Probably more will show up)

Reuse existing helpers like `setOLSubLayers` and `getAllLayerIdsInGroup`.

This will be added as a standalone API only. Existing LayerSwitcher, URL hash logic, and other features will not be migrated.

jesade-vbg:
Here is a generated testing procedure that I found very useful.

### LayerControl API — Manual Testing

Quick smoke-test guide for the centralized layer toggle API
(`LayerControlModel`, exposed on `window.hajkPublicApi.layerControl`).

Run all commands in the **browser console** while the client is running
(`cd apps/client && npm run dev`).

After every toggle, verify that the **LayerSwitcher checkbox/radio updates
automatically** — the UI mirrors OpenLayers state via `propertychange`
listeners, so no manual refresh is needed.

## 1. Grab the API

```js
const lc = window.hajkPublicApi.layerControl;
const map = window.hajkPublicApi.olMap; // the OL map, for inspecting ids below
```

## 2. Find ids to test with

A layer's id is its OpenLayers `name` property. Note `getAllLayers()` already
returns a plain array (no `.getArray()` needed).

```js
// All layers with their type
map.getAllLayers().map((l) => [l.get("name"), l.get("layerType")]);

const L = map.getAllLayers();
L.filter((l) => l.get("layerType") === "group").map((l) => l.get("name")); // WMS group layers
L.filter((l) => l.get("layerType") === "base").map((l) => l.get("name"));  // backgrounds
```

For a LayerSwitcher tree-group **folder** id, probe ids from the LayerSwitcher UI
groups — `lc.showGroup(id)` warns if the id is not a folder.

## 3. Normal layer

```js
lc.showLayer("<id>");
lc.hideLayer("<id>");
lc.toggleLayer("<id>");
lc.setLayerVisibility("<id>", true); // boolean form of show/hide
lc.layerIsVisible("<id>"); // -> true / false
```

## 4. WMS group layer + its sublayers

A WMS group layer (`layerType === "group"`) is one OpenLayers layer with several
**sublayers**. The sublayers are NOT standalone layers, so you must always
address them via the parent group layer id — `showLayer("<subLayerId>")` /
`toggleLayer("<subLayerId>")` will fail (sublayer is not a real OL layer).

```js
// Whole group layer (all sublayers on/off)
lc.showLayer("<groupId>");                                   // all sublayers
lc.showLayer("<groupId>", { subLayers: ["sub1", "sub2"] });  // subset
lc.toggleLayer("<groupId>");

// Individual sublayers (the nested checkboxes)
lc.showSubLayer("<groupId>", "sub1");
lc.hideSubLayer("<groupId>", "sub1");
lc.toggleSubLayer("<groupId>", "sub1");
lc.setSubLayerVisibility("<groupId>", "sub1", true); // boolean form
lc.setSubLayers("<groupId>", ["sub1", "sub3"]);      // exact visible set
lc.subLayerIsVisible("<groupId>", "sub1");           // -> true / false
```

### Composite `"groupId:subLayerId"` shorthand

`showLayer` / `hideLayer` / `toggleLayer` / `layerIsVisible` also accept a single
`"groupId:subLayerId"` string and route it to the matching sublayer method.
(A real layer whose name happens to contain `:` still takes precedence, so this
never breaks existing ids.)

```js
lc.toggleLayer("<groupId>:sub1");          // === lc.toggleSubLayer("<groupId>", "sub1")
lc.showLayer("<groupId>:sub1");            // === lc.showSubLayer(...)
lc.hideLayer("<groupId>:sub1");            // === lc.hideSubLayer(...)
lc.layerIsVisible("<groupId>:sub1");       // === lc.subLayerIsVisible(...)
```

Example for the "Trafiksamordning linje" group layer (`70d0ft`) and its
"Påbörjad" sublayer:

```js
// Either of these work and are equivalent:
lc.toggleLayer("70d0ft:trafiksamordning_linje_paborjad");
lc.toggleSubLayer("70d0ft", "trafiksamordning_linje_paborjad");

lc.layerIsVisible("70d0ft:trafiksamordning_linje_paborjad");
```

To list a group layer's sublayer ids:

```js
map.getAllLayers()
  .find((l) => l.get("name") === "70d0ft")
  .get("allSubLayers");
```

## 5. Base / background layer (exclusive radio)

Showing a base layer hides all other base layers (radio behavior), syncs the
LayerSwitcher radio, and repaints the `div#map` background.

```js
// Any base layer by id. showBackground warns if the id is not a background.
lc.showBackground("<baseId>");
lc.showLayer("<baseId>"); // equivalent low-level call

// Built-in "special" backgrounds — named shortcuts, no magic ids needed.
lc.showBackgroundWhite(); // "-1" -> map canvas turns white (#fff)
lc.showBackgroundBlack(); // "-2" -> map canvas turns black (#000)
lc.showBackgroundOSM();   // "-3" -> OpenStreetMap tiles

// Equivalent to the raw ids:
lc.showLayer("-1");
lc.showLayer("-2");
lc.showLayer("-3");
```

Verify the result (radio updates automatically, and the canvas repaints):

```js
lc.layerIsVisible("-1");                              // -> true / false
document.getElementById("map").style.backgroundColor; // -> "rgb(255, 255, 255)" after white
map.getAllLayers()
  .filter((l) => l.get("layerType") === "base")
  .map((l) => [l.get("name"), l.get("visible")]);      // only one base visible
```

`showBackground` is guarded — passing a non-background id warns and does nothing:

```js
lc.showBackground("<normalOrGroupLayerId>"); // console warning, no change
```

## 6. LayerSwitcher tree-group folder (operates on every leaf)

```js
lc.showGroup("<folderId>");
lc.hideGroup("<folderId>");
lc.toggleGroup("<folderId>");
lc.setGroupVisibility("<folderId>", true);
lc.groupIsVisible("<folderId>"); // -> true / false (all leaves visible)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Client: Add centralized LayerControlModel API (showLayer/hideLayer/toggleLayer/layerIsVisible) for toggling layers by id, exposed on the AppModel and `window.hajkPublicApi`, plus `useLayerVisibility`/`useLayerState` React hooks.
 - Client + Admin: LayerSwitcher - Added a new admin setting "Visa teckenförklaring direkt" that forces the legend to be expanded by default in the layer details view, so users don't have to click the legend button. [#1838](https://github.com/hajkmap/Hajk/issues/1838)
 - Client + Admin: DocumentHandler - Added a "Direct Print" setting that prints the currently active document directly without showing the document selection dialog [#1773](https://github.com/hajkmap/Hajk/issues/1773)
 - Client: Infoclick - Functionality to hide links that point to non-existing resources [#1804](https://github.com/hajkmap/Hajk/issues/1804)
--- a/apps/client/src/components/App.jsx
+++ b/apps/client/src/components/App.jsx
@@ -488,6 +488,7 @@ class App extends React.PureComponent {
       .createMap()
       .addSearchModel()
       .addLayers()
+      .addLayerControl() // Layer control must be added after the layers
       .addAnchorModel() // Anchor model must be added after the layers
       .loadPlugins(this.props.activeTools);
 
@@ -562,6 +563,7 @@ class App extends React.PureComponent {
     window.hajkPublicApi = {
       ...window.hajkPublicApi,
       olMap: this.appModel.map,
+      layerControl: this.appModel.layerControl,
       dirtyLayers: {},
     };
 
--- a/apps/client/src/constants/backgroundLayers.js
+++ b/apps/client/src/constants/backgroundLayers.js
@@ -1,0 +1,13 @@
+/**
+ * Ids for the "special"/non-server background layers that Hajk creates client
+ * side (see `models/appModel/backgroundLayers.js`). They are real OpenLayers
+ * layers added to the map, but the white/black ones only render via the
+ * `div#map` background color rather than any tiles.
+ *
+ * The id equals the layer's OpenLayers `name` property.
+ */
+export const BACKGROUND_LAYER_IDS = {
+  WHITE: "-1",
+  BLACK: "-2",
+  OSM: "-3",
+};
--- a/apps/client/src/hooks/useLayerVisibility.js
+++ b/apps/client/src/hooks/useLayerVisibility.js
@@ -1,0 +1,47 @@
+import { useSyncExternalStore, useCallback } from "react";
+
+/**
+ * Subscribe to a layer's visibility and re-render automatically whenever it
+ * changes, no matter who toggled it (LayerSwitcher, URL hash, another plugin or
+ * the LayerControl API). Also accepts a LayerSwitcher tree-group folder id, in
+ * which case the value is `true` only when every leaf layer is visible.
+ *
+ * @param {import("../models/LayerControlModel").default} layerControl
+ * @param {string} id Layer id (OpenLayers `name`) or tree-group folder id.
+ * @returns {boolean} Current visibility.
+ */
+export function useLayerVisibility(layerControl, id) {
+  const subscribe = useCallback(
+    (cb) => layerControl.subscribe(id, cb),
+    [layerControl, id]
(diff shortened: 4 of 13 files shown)


jesade-vbg:
Fixed some eslint issues in Favorites etc when I passed by.

jesade-vbg:
Added OSM Vector support to const etc and included info in AGENTS.md.

jacobwod (changes requested):
Very neat idea. I like the layerControl API on hajkPublicApi a lot.

A few suggestions:
- consider renaming some of the methods. E.g. I think that `isLayerVisible()` and `isGroupVisible` would be more inline with how other APIs call similar functions (e.g. OpenLayer's `isVisible`). 
- the API seems to expose a lot of private methods of the class (`#globalObserver`, etc). If possible we should try to restrict it so the public API stays clear. Might not be a huge problem and they're correctly prefixed with `#`, but if we can find a solution (perhaps some `structuredClone` that strips out those parts before appending on `hajkPublicApi`, or perhaps TS with real `private` indicators for class members?) would solve it.

--- 

Toggling layers with auto-hiding works as expected for background layers (even when not using the `show*Background` methods:

Image

But it fails for exclusive groups (the recently added `exclusive` property of regular Hajk groups, that render as radio buttons):

Image

This could be an oversight as the exclusive groups functionality has been added very recently, but please have a look. 

If I'm using the API functions in a wrong way: let me know. If it's not meant to toggle background layers and/or exclusive group layers using the regular `showLayer`: perhaps there should be a check that throws a warning to the console and explains that?

PR #1874, opened 2026-07-10
Feature/1873 infoclick value mapping filter

### LINKED ISSUE #1873: Infoclick: Add value mapping filter
When using `{click:zoom}` in an infoclick external link to pass the current zoom index to another instance, that instance sets its view to the same index. If the `resolutions` arrays are different between the instances, the same zoom index can result in a different map scale.

The suggestion is to add a filter that maps an attribute value to a configurable output.

Example for mapping zoom index between instances with different `resolutions` arrays:
`{click:zoom|map('1:2,2:3')}`

This issue is similar to issue #1865.

linusfj:
Closes #1873.

### DIFF
--- a/apps/client/src/components/FeatureInfo/FeaturePropsFilters.js
+++ b/apps/client/src/components/FeatureInfo/FeaturePropsFilters.js
@@ -23,8 +23,8 @@ class PropFilters {
     } else {
       value = this.properties[key];
 
-      if (typeof value === "boolean") {
-        // We must convert to string to avoid boolean comparisons
+      if (typeof value === "boolean" || typeof value === "number") {
+        // We must convert to string to avoid boolean or numeric comparisons
         // Everything should be stringified
         value = String(value);
       }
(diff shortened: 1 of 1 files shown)


jesade-vbg:
Looks good overall. One small concern: since the mapping string is split on commas, a mapped value can't itself contain a comma, e.g. "1:Damaged, needs repair". Is that an intentional limitation we should document, or should we support escaping/quoting?

Also, invalid mapping pairs without ":" are currently silently ignored. Would it be better to throw/log a warning so configuration mistakes are easier to spot?


And I have another question, If its just adding + to zoom level which it usually is, then I would prefer a simple
`{click:zoom|add(1)}`
instead of 
`{click:zoom|map('0:1,1:2,2:3,3:4,4:5,5:6,...')}`

So I the purpose is only Add, then you should just add an add filter instead.

Hallbergs (approved):

PR #1881, opened 2026-08-17
Feature/map follow location

### LINKED ISSUE #1875: Follow position
It would be nice to give a user the option to have the location plugin not only show the current position, but also continuously recenter the map to the current position.

Would be useful when using the map while moving without the need to zoom/pan until the position tracking is stopped.

This toggle-button should be configurable to show by the admin aswell.

As a proper tv-chef I have already started a branch with a suggested solution that id appreciate some yay or nays on . [feature/map-follow-location](https://github.com/hajkmap/Hajk/tree/feature/map-follow-location)

michaelsund:
Adds a snackbar when tracking is active, the snackbar toggles the "follow" capability and recenters the map on the current position when a new position event is recieved.

### DIFF
--- a/apps/admin/src/views/tools/location.jsx
+++ b/apps/admin/src/views/tools/location.jsx
@@ -46,6 +46,7 @@ class ToolOptions extends Component {
         width: tool.options.width,
         height: tool.options.height,
         visibleAtStart: tool.options.visibleAtStart,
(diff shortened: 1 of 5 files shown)


jacobwod (approved):
Super, nice job!

Issue #1675, opened 2025-08-25
New Editing Tool for HAJK (Traditional + Quick Capture)


Sunix71:
**Summary**
Introduce a brand-new (greenfield) editing tool for HAJK to make the experience simpler to maintain and mobile-friendly. The solution will have two user-facing parts: a **Traditional Editing** tool (form/list-based) and a **Quick Registration** tool (few taps, field-first). We are still investigating whether **OGC API Features** or **WFS(-T)** will be used.

**Goals & Foundations**
* Rebuild from scratch; the current editor is too complex to safely evolve.
* Reuse existing HAJK components wherever possible (especially **Sketch** for geometry).
* Persist edits via the configured **map server endpoints**.
* Offline mode is not in scope for the first phase, but the architecture must keep offline support feasible later.
* Nice-to-have: online-first sync that automatically uploads when connectivity returns.

**Traditional Editing Tool**
A form/list-driven editor that works equally well on desktop and mobile. Users create and edit geodata **through configured editing services on the map server**; changes are persisted server-side and reflected in the database. The editing map should expose layers and tool buttons consistent with the rest of HAJK and support multiple editing services per map. Consultants with HAJK logins must be able to edit. Nice-to-have: real-time presence of users/devices (e.g., via websockets).

**Quick Registration Tool**
Mobile-first tool for highly repetitive field workflows (e.g., confirming a lawn was mowed), citizen reports with GPS, and rapid crisis mapping (e.g., flood positions + photo). Minimize taps; some attributes auto-filled by admin settings.

**Technical Notes**
* Backends: PostgreSQL and MS SQL Server (via the map server).
* Standards: **WFS-T or OGC API Features** (under evaluation); aim to align with OGC API Features long term.
* Real-time user/device presence could be implemented with websockets.
* Multi-geometry tables (MultiLineString/MultiPoint/MultiPolygon) must be handled: single geometries should be automatically wrapped to multi on save.
* Consider admin-selectable GPS receiver (device GPS vs RTK-enabled receiver) for centimeter-level accuracy (nice-to-have).
* Nice-to-have: add object/file/image directly during editing; add vertices at line ends and reverse line direction; switch a layer from WMS to WFS within HAJK.

Sunix71:
_Edit by @jacobwod: just fixed formatting to make reading easier. No other changes to @Sunix71's post._

**Status update and decisions taken, requesting approach confirmation before PR**

[...]

PR #1882, opened 2026-09-01
AttributeEditor: new editing tool for WFS-T layers (#1675)

Sunix71:
**Summary**
Implements Editing tool from #1675, a new editing plugin (AttributeEditor) for WFS-T layers, following the approach confirmed in the issue.

**Client:** new AttributeEditor plugin with table and form editing modes: bulk attribute editing, per-column faceted filtering, duplicate/split/merge of features, Excel export and mobile support. 
Geometry work integrates with the existing Sketch plugin through a small documented CustomEvent bus (editBus).

**Admin:** new tool view for configuring editing services (writes standard layers.json/map_1.json structures).

**Backend:** new /api/v2/ogc/wfst endpoints serving WFST layer metadata (with an optional WFS_ALLOWED_HOSTS allowlist), plus hardening of the generic proxy's identity handling (PROXY_NAME_FORWARD_VALIDATED_USER / _STRIP_AUTH / Basic auth injection with startup warnings). 
The WFS/WFS-T traffic itself goes directly from the browser to the map server, so existing per-user authorization (Kerberos/NTLM) applies. Happy to split the proxy-hardening portion into a separate PR if preferred.

**Testing**
Tested continuously against real QGIS Server and GeoServer instances (PostgreSQL and Oracle-backed layers), by me and several testers in separate environments. Covered flows: create/edit/delete including bulk edits, drawing, move, rotate, split and merge of geometries, multi-geometry wrapping on save, column filtering including boolean and datetime facets, save/discard/undo flows, service switching with unsaved changes, KML/GPX import into the editing layer, both proxied and direct layer URLs, and the proxy identity hardening (curl matrix including the Expect: 100-continue bypass).

**Known debt and follow-ups (a follow-up cleanup PR is planned)**

- Consolidation: the save flow, selection state, undo stacks and id-lookup structures each exist in 2-4 copies kept in sync manually. Working and individually verified, but mapped for consolidation (single save hook, single selection set, unified undo journal).
- Deliberate behaviors: a partially failed WFS-T save commits locally and reloads server truth. Transactions are atomic, and keeping pending edits would duplicate inserts on retry, so the failure messaging was improved instead. The Sketch window closes when the AttributeEditor opens at the same position. This is known upstream BaseWindowPlugin behavior (maintainers are aware), and the workaround is configuring the windows at different positions.
- Hardening follow-ups: WFST layer metadata is not yet washed per layer against visibleForGroups (the map config is washed, and the gap is noted as a TODO in code). QGIS Server detection: fixed in this PR. The admin WFST form now exposes a serverType dropdown, stored on the layer config and honored before the URL substring fallback (which remains for unconfigured layers). Custom idField edge cases remain mapped for the cleanup PR.
- Remaining lint: two hook-level items under React's new lint rules are suppressed with documented rationale pending the cleanup PR. One pre-existing warning in an untouched upstream file (BufferView) is left as-is.


jacobwod:
Hi @Sunix71! This looks good in general, but please allow us to do a proper review and testing. The goal is to merge this before the next release (4.4) so this PR makes it to the release candidate test round and gets throughout testing. To prepare for the merge: can you please take a look at this branch and merge the most recent `develop`? Thanks.

Sunix71:
Hi @jacobwod, thanks! Develop is now merged into the branch.

I checked the new CommandPalette against the AttributeEditor and pushed two small follow-ups: the editor's vector layer is now tagged as a system layer, so the palette, quick access and favorites treat it the right way, and it got an explicit zIndex of 5000.

Ready for review and testing whenever you are.

Hallbergs (dismissed):

jesade-vbg (dismissed):

Sunix71:
Merged the latest develop again (CHANGELOG was the only conflict). The merge dismissed your approvals as stale, so a fresh approval is needed, sorry for the extra round!

Since a new approval was needed anyway, I included one small fix I found while smoke testing the merge, in a separate commit: the AttributeEditor layer had no caption, so it showed up as an empty row under system layers in the draw order view. One line, no other changes.

A small tip for the command palette (I left its file alone since it's not part of this PR): the tool is listed as "Attributeeditor" because attributeeditor isn't in PLUGIN_TITLES. Adding attributeeditor: "Attributredigerare" there would give it its Swedish name.

jesade-vbg (approved):

jesade-vbg:
Great work @Sunix71 ! This will be super.

PR #1891, opened 2026-09-11
Command palette for common operations in Client

### LINKED ISSUE #1867: Client: Add a Command Palette to perform various actions
_Update_
This took off well, here's the current list of commands:

- Open a tool
- Toggle layer visibility
- Switch background layer
- Preset links (fly to-functionality)
- Various search options (polygon, radius, etc)
- Use query term in search box (regular search) as well as clear the search box
- Switch theme (light/dark)
- Close all windows
- Hide all layers

I think that one major thing to fix is:
- add guarding to the various options. E.g. if `search` plugin is not configured in map config, we should not show any of the search-related commands
- refactor so that layer switching uses the same functionality as the regular layer switcher and calls that via the `globalObserver` (that's how PresetLinks commands work now - they just call the regular PresetLinks rather than duplicating its functionality).

Feel free to contribute with code and ideas for commands to add!

---
_Original issue_
Imagine having a command palette in Hajk that would allow power users to quickly navigate the app. 

I got the idea from VSCode's command palette and I think we should implement something very similar. 

- It should activate on <kbd>Cmd</kbd> + <kbd>K</kbd>, as it seems to have become some kind of standard for opening a command palette (go ahead and google it)
- It should show a list of possible actions and filter the list as user starts typing
- If other shortcuts are available for the most common actions, those should be printed (just as in the screenshot below)

Image

I got this idea when i noticed that I often must reach out for the trackpad only to focus the search field. So focusing the search field would be one obvious use case. 

Another could be controlling plugins. Imagine this flow:
1. <kbd>Cmd</kbd> + <kbd>K</kbd>
2. Type "op" which filters the command list and focuses on the command called "Open tool…"
3. Press <kbd>Enter</kbd>
4. Now a list of available Hajk tools shows up in the command palette and user can either navigate with up/down arrows or filter by typing, e.g. "lay"
5. This filters down and focuses on the one tool that matches the input: LayerSwitcher
6. Press <kbd>Enter</kbd> which completes the command palette action, closes it and opens LayerSwitcher

OK, this seems like a lot in text but try it in a app that has a command palette. Takes 0.5 seconds at most. I really think this could be useful for power users.

_This is slightly related to, but possibly even superseeds #257._

jacobwod:
See #1867 for details.


michaelsund:
Added guards to not show search suggestions if the search-plugin isn't enabled in the backend.

michaelsund (approved):

Hallbergs (approved):

